PR #22, opened 2021-03-25
Adds a virtual init method to flow nodes

sturcotte06:
This adds a virtual init method that is analoguous to the cleanup method. It makes it easier to declare resources.


MothDoctor:
Hey, it seems to be a nice idea! :)

Accepted that, but didn't pull the original code. Decided to name method ```InitializeInstance```. I also liked calling methods ```Init```, but that's quite bad when you're trying to find by the global search, both in editor or IDE. 

Thanks for another contribution! :)

PR #29, opened 2021-04-03
Adds a flow graph component and implement world settings using this component for re-usability

sturcotte06:
I was thinking about it, and I think it might be better to have a component that owns a flow asset and registers it automatically to the subsystem. This way, we can re-use the logic within the world settings anywhere a flow asset is needed. What do you think?

### DIFF
--- a/Config/BaseFlow.ini
+++ b/Config/BaseFlow.ini
@@ -1,3 +1,4 @@
 [CoreRedirects]
 +ClassRedirects=(OldName="/Script/Flow.FlowNode_CustomEvent",NewName="/Script/Flow.FlowNode_CustomInput")
 +PropertyRedirects=(OldName="FlowAsset.CustomEvents",NewName="CustomInputs")
++PropertyRedirects=(OldName="FlowWorldSettings.FlowAsset",NewName="FlowGraph.FlowAsset")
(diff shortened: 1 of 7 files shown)


MothDoctor:
That looks interesting! I gonna pull this in :)

I'm thinking about 2 changes here, gonna check it with the internal project.
1. Calling StartRootFlow (and End too) from the distinct methods instead of BeginPlay/EndPlay. Actually, this would solve an issue where the Root Flow shouldn't start together with the world's Begin Play. If some game-specific has to be met before starting the story/campaign. 
These methods would be still called by default on BeginPlay/EndPlay

2. Need to meditate a bit, think if we can have a less confusing name for the component. "Flow Graph Component" might be confusing if used instantiate Flow Asset (graph might be one of many asset elements). Someone might assume that component has something to do with executing graph logic. For now, I'm thinking about the "Root Flow Component", the first one to try. This name should help with getting used to "Root Flow" term.
"Flow Spawner Component" or "Flow Subsystem Component" were other candidates, but not sure these names clearly indicate the purpose of the component.

MothDoctor:
Mhm... Actually, it might be much better NOT to create an additional component class. No matter how we would name it, it would be confusing for users. Especially if the given actor would have both components, i.e. Player Controller with Flow Component and "Flow Graph Component".
Role of Flow Component is already to communicate between AActor and the Flow system :)

This feature can be simply added Flow Component itself :)
As bonus, owner can be tagged on the same component.

MothDoctor:
@sturcotte06 your idea got accepted, just expanded it a bit :)
Thanks for another contribution :)

PR #31, opened 2021-04-05
Add MultiGate node which replicates behavior similar Blueprint node

BorMor:
MultiGate has Random, Loop and StartIndex properties

Any feedback are appreciated

### DIFF
--- a/Source/Flow/Private/Nodes/Route/FlowNode_ExecutionMultiGate.cpp
+++ b/Source/Flow/Private/Nodes/Route/FlowNode_ExecutionMultiGate.cpp
@@ -1,0 +1,110 @@
+#include "Nodes/Route/FlowNode_ExecutionMultiGate.h"
+
+#include "Misc/StringBuilder.h"
+
+const FName UFlowNode_ExecutionMultiGate::ResetInputName = TEXT("Reset");
+
+UFlowNode_ExecutionMultiGate::UFlowNode_ExecutionMultiGate(const FObjectInitializer& ObjectInitializer)
+	: Super(ObjectInitializer)
+{
+#if WITH_EDITOR
+	Category = TEXT("Route");
+	NodeStyle = EFlowNodeStyle::Logic;
+#endif
+
+	InputNames.Empty();
+	InputNames.Add(DefaultInputName);
+	InputNames.Add(ResetInputName);
+	
+	SetNumericalOutputs(0, 1);
+}
+
+void UFlowNode_ExecutionMultiGate::ExecuteInput(const FName& PinName)
+{
+	if (PinName == ResetInputName)
+	{
+		Finish();
+		return;
+	}
+
+	if (Completed.Num() == 0)
+		Completed.Init(false, OutputNames.Num());
+
+	if (!Completed.Contains(false))
+		return;
+
+	const bool bUseStartIndex = !Completed.Contains(true) && Completed.IsValidIndex(StartIndex);
+	
+	if (bRandom)
+	{
+		int32 Index = INDEX_NONE;
+		if (bUseStartIndex)
+			Index = StartIndex;
+		else
+		{
+			TArray<int32> AvailableIndexes;
+			AvailableIndexes.Reserve(Completed.Num());
+			
+			for (auto It = Completed.CreateConstIterator(); It; ++It)
+				if (*It == false)
+					AvailableIndexes.Add(It.GetIndex());
+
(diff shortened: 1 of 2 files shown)


MothDoctor:
It looks tasty, definitely gonna be useful in many projects! :)

For now, I only had time to view code on the web, so I noticed mostly the coding style that wouldn't exactly fit plugin style.
- Plugin's code follows Unreal coding standard, so here I'd like to ask you to add braces to If and For blocks :)
https://docs.unrealengine.com/en-US/ProductionPipelines/DevelopmentSetup/CodingStandard/index.html

- I would avoid using advanced C++ things if there's no noticeable gain. This is to make it easier for less experienced programmers - which are very likely to use Flow Graph since it's a high-level scripting system. 
In this PR I would replace `CreateConstIterator()` with the standard loop. And using actual types instead of `auto`.
This piece might be utterly confusing for many.
```
for (auto It = Completed.CreateConstIterator(); It; ++It)
        if (*It == false)
	    AvailableIndexes.Add(It.GetIndex());
```

I wonder if using TStringBuilder<128> has a noticeable performance advantage over FString?
If yes, it's something new to me and I should get familiar with it :)
If not, I'd consider using plain FString for consistency with the rest of the codebase, although I admit this goes a bit into an area of nitpicking ;)

I guess that we could initialize `Completed` array in the constructor? This and `NextOutput` would need to become UPROPERTY anyway. It's because I gonna implement save/load support for graphs soon :)
```
if (Completed.Num() == 0)
		Completed.Init(false, OutputNames.Num());
```

That's it for now. I'd actual testing on another day :)

BorMor:
> I wonder if using TStringBuilder<128> has a noticeable performance advantage over FString?
Yes, every time then you try to append string and allocate new memory and copy string into it and then freed previous memory
TStringBuilder pre-allocate memory for string operations, it allow make fewer unnecessary allocations.

> I guess that we could initialize Completed array in the constructor? 
I don't know how many outputs has node in constructor, because user may add them :)
So I initialize array on first run, and use it as "Initialized" flag.

sturcotte06:
> > I wonder if using TStringBuilder<128> has a noticeable performance advantage over FString?
> 
> Yes, every time then you try to append string and allocate new memory and copy string into it and then freed previous memory

Not really, you would have the same behaviour with:

```cpp
FString String;
String.Reserve(128);
```

`FString` are mutable, which means there is no re-allocation if you append to it. There is only a re-allocation if the buffer size is not big enough to hold the string.

sturcotte06:
> > I wonder if using TStringBuilder<128> has a noticeable performance advantage over FString?
> 
> Yes, every time then you try to append string and allocate new memory and copy string into it and then freed previous memory

Not really, you would have the same allocation with:

```cpp
FString String;
String.Reserve(128);
```

`FString` are mutable, there is only a re-allocation if the current buffer is not big enough to hold the string.

MothDoctor:
So, this would be the best way of writing this, @sturcotte06? :)

```c++
FString String;
String.Reserve(128);

Builder += TEXT("StartIndex: ");
Builder += FString::FromInt(StartIndex);
```

sturcotte06:
> So, this would be the best way of writing this, @sturcotte06? :)
> 
> ```c++
> FString String;
> String.Reserve(128);
> 
> Builder += TEXT("StartIndex: ");
> Builder += FString::FromInt(StartIndex);
> ```

Yes I think so, I'm not a 100% sure as I didn't test it, but my educated guess would be:

```cpp
// Will call operator+=(TCHAR*)
Builder += TEXT("StartIndex: ");

// Will create a temporary FString and call operator+=(const FString&)
Builder += FString::FromInt(StartIndex);
```

MothDoctor:
> EDIT: No it's not the best way, the best would be with `Appendf` since it does not create any temporaries:
> 
> ```c++
> String.Appendf(TEXT("StartIndex: %d"), StartIndex);
> ```

OK, I gonna switch to use Append. Totally forgot about it, thanks! :)

MothDoctor:
@BorMor Finally pulled that in :)

Tested it, no issues found. Only it was initially confusing that `bLoop` is needed to execute gates again. Confusing because I didn't use Gate nodes for a long time. So I added a comment to this flag.
Also made cosmetic changes to code, so it looks closer to other nodes. Plus little things, like adding spaces (node name, node description) and this icon in the top corner :)

UE4Editor_lCTB4nfkxd

Thanks for contribution, should be very useful :)

PR #43, opened 2021-04-29
AutowireNewNode

IceRiverr:
When creating a new node, connect to the new node.


MothDoctor:
@IceRiverr Fix accepted, thanks for your contribution! 
And fixing what I totally forgot to fix ;)

PR #44, opened 2021-05-04
Extra Sequencer node features (Flow Actor Bindings and Aspect Ratio constraint override)

LunaRyuko:
[(Video demo)](https://youtu.be/tkMmW0hazzw)
This PR adds extra properties to the Sequencer node that are available on Level Sequence Actors, like Object Bindings and Aspect Ratio override stuff.
UE4Editor_fAejCAxsVl

NOTE: I made the Override Aspect Ratio Axis Constraint disabled by default in contrast to the original level sequence actor, because my project uses Maintain FOV on Y-Axis by default.


MothDoctor:
@LunaRyuko Finally decided not to merge this pull request. This node was meant to be the simple base for playing Level Sequence. 
Things like changing the Aspect Ratio belong more to subclassed `Cinematic` node.
Binding actors might be performed differently in different games/systems. so adding it to the base node might be confusing.

I'd definitely might come back to this PR when working on examples, i.e. "Play Cinematic" mode :)

MothDoctor:
Hey @LunaRyuko after all decided to expose Camera Settings on the node, my approach has been corrected.

Couldn't merge your code directly, simply implement it my way.
https://github.com/MothCocoon/FlowGraph/commit/9408bfd5d7e8aea1ac276ad2b79f829865f97cb9

In my spare time I'd look if it would be fine to expose other Level Sequence Actor parameters.
Thanks for PR! :)

PR #49, opened 2021-05-31
fix UFlowAsset::ClearInstances

Mephiztopheles:
If you load a savegame and exit the game afterwards you can get an exception because activeInstances is empty. I've sticked to your style and added a check like in FlowSubSystem::Deinitialize


MothDoctor:
@Mephiztopheles Thanks for the fix, pulled that in! :)

PR #51, opened 2021-08-14
Fixed Engine parsing Plugin version error while cooking

Solessfir:
This warning appears when you try to Build a Project:
`LogPluginManager: Warning: Engine version string in .../Plugins/FlowGraph/Flow.uplugin could not be parsed ("5.0")`


MothDoctor:
Thanks for the fix @Solessfir :)

PR #53, opened 2021-09-20
Allow hierarchical identity searches

iknowDavenMC:
When searching for flow components with the requested identity, use MatchesTag on each registered flow component instead of MultiFind on the entire container to allow for hierarchical identity searches (e.g. a flow component with identity Identity.General.Specific can be found by searching for Identity.General)

### DIFF
--- a/Source/Flow/Private/FlowSubsystem.cpp
+++ b/Source/Flow/Private/FlowSubsystem.cpp
@@ -7,6 +7,7 @@
 #include "FlowSettings.h"
 #include "Nodes/Route/FlowNode_SubGraph.h"
 
+#include "Algo/Transform.h"
 #include "Engine/GameInstance.h"
 #include "Engine/World.h"
 #include "Misc/FileHelper.h"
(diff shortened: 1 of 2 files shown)


MothDoctor:
Just wanted to comment here, as PR waits for a long time.

I hesitate to merge mostly because I'm afraid that would degrade performance for all the searches, as it would no longer utilize the fixed cost of querying hash container? Or I am wrong on this matter?

iknowDavenMC:
Based on the comment above `TMapBase`'s definition, `TMultiMap` lookup complexity is O(1), whereas iterating over all elements with `Algo::TransformIf` has a complexity of O(n) (where n is the number of elements in `FlowComponentRegistry`). Therefore, you are correct; this change will decrease the worst case performance of all searches.

That being said, I think that allowing for hierarchical searches is worth the cost.
* Hierarchy is one of the main characteristic of `FGameplayTag`s. Therefore, as a user, my default expectation is for systems that use `FGameplayTag`s to leverage hierarchy.
* Hierarchical searches are useful. This change under review was born out of an actual need on our project.

However, perhaps this change would need to be part of a larger move towards hierarchical use of flow identity tags (I assume there are other parts of the plugin that would need similar changes), or perhaps we would need to expose hierarchical mechanics like this one as advanced options as not to negatively impact those who don't need them.

What do you think?

MothDoctor:
> However, perhaps this change would need to be part of a larger move towards hierarchical use of flow identity tags (I assume there are other parts of the plugin that would need similar changes), or perhaps we would need to expose hierarchical mechanics like this one as advanced options as not to negatively impact those who don't need them.
> 
> What do you think?

Just wanted to drop info here that I'd like to look into it during February, so it could be included in the plugin version.
Indeed, I hope to proving your extension as advanced option, so users would be aware of performance trade-off :)

MothDoctor:
@iknowDavenMC finally implement search by non-exactly matching tags. It's exposed via `bExactMatch` bool, the same way as blueprint FGameplayTagContainer methods expose it.

Added information that `bExactMatch` parameter is an expensive option to use.

Thanks for proposing this change in the first place!

PR #54, opened 2021-09-25
Add circuit connection drawing.

ryanjon2040:
Added circuit line drawing for Flow Graph.


MothDoctor:
@ryanjon2040 I pulled that change :)
For now, I only switched the default connection drawing style to the... default one ;)

Don't be shy to send updates to the drawing algorithm, if you'd work on it more. This PR got automatically closed upon merge.
What I observed from the first inspection
- Sometimes perfectly straight wire doesn't want to be drawn as straight, perhaps it's because some nodes got pins "misaligned" by a pixel. 
- Wires might look bad if zoom out, or the wire goes up and left.

PR #55, opened 2021-10-01
Implement graph tooltip.

ryanjon2040:
Added an option to show sub graph as tooltip. Toggleable from settings.


ameaninglessname:
The comment bubble can not scale when the event graph scales, could we fix that or provide a PR to the engine or just use the "comment box (hold C + left click)"

PR #62, opened 2021-10-22
Execute any pin in a Flow Graph externally from Bluprint

ryanjon2040:
From any Blueprint you can call this node to execute a specific input pin in Flow Graph. Keep in mind that this node *does not* respect *bAllowMultipleInstances*.


MothDoctor:
@ryanjon2040 Hi, I'm closing down this PR as feedback hadn't been addressed, and there are conflicts after so long time.
Don't hesitate to open a new PR if you'd find worth to refactor it :)

PR #65, opened 2021-10-26
Make automatically placed Start node as ghost node.

ryanjon2040:
Small cosmetic tweak. Automatically placed start mode will be placed as ghost node 👻 

Flow_Disabled_Node

### DIFF
--- a/Source/FlowEditor/Private/Graph/FlowGraphSchema.cpp
+++ b/Source/FlowEditor/Private/Graph/FlowGraphSchema.cpp
@@ -72,6 +72,7 @@ void UFlowGraphSchema::CreateDefaultNodesForGraph(UEdGraph& Graph) const
 {
 	// Start node
 	UFlowGraphNode* NewGraphNode = FFlowGraphSchemaAction_NewNode::CreateNode(&Graph, nullptr, UFlowNode_Start::StaticClass(), FVector2D::ZeroVector);
+	NewGraphNode->MakeAutomaticallyPlacedGhostNode();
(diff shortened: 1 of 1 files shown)


MothDoctor:
@ryanjon2040 I'm not sure about this change. We actually want this node to be very visible in the standard Flow Graph as it must be used.

I understand the need for it, If you'd custom Flow Asset type where using Start event might be optional. Perhaps it would make sens to make this ghosting based on the type of the Flow Asset? What do you think?

ryanjon2040:
Hi @MothDoctor I think the ghosting won't be a problem since it is just a visual change similar to how `Begin Play` in Blueprint but at the same time what you said make sense too because the `Start` node must indicate that no matter what, flow starts from here. How about you ask `@everyone` in Discord maybe? Or maybe you can make this for Custom Inputs so when they initially place it in graph it will be the ghost node until something is connected?

MothDoctor:
Oh, I was pretty sure I replied here a long time ago.

@ryanjon2040 What if we would Flow Asset property to control this behavior. By default, Start wouldn't be ghosted, but you could easily set it for your custom Flow Asset class? :)

ryanjon2040:
@MothDoctor Ya probably a setting in Flow Asset could be useful 🙂

MothDoctor:
@ryanjon2040 PR has been merged in.
I implement my original proposal. This way a default behavior isn't changed, but might simply be modified by subclass `UFlowAsset`.

Thanks for contribution :)

PR #67, opened 2021-11-02
Node now have tick event.

ryanjon2040:
Flow nodes now have Tick event

.


MothDoctor:
I deliberately didn't support Tick in nodes. It would lead to applying bad code architecture by a lot of users instead of embracing event-based architecture, they would simply avoid using a lot of the benefits of Flow Node design.

Obviously, it's easy to implement `FTickableObject` for any native Flow Node.

For this reason, I'm not gonna accept this PR... for now. Perhaps I would change my mind later, now I wanted to explain why this wouldn't be simply pull in.

ryanjon2040:
Got it! Understood and what you said makes sense. Thank you :slight_smile:

PR #71, opened 2021-11-12
Fix user can delete Start node.

ryanjon2040:
This PR fixes deleting of `Start` node. It is also possible to set this setting per `Flow Node` instead of `Flow Graph Node` eliminating the need of creating additional `Flow Graph Node` classes if you don't want the node to be deleted or duplicated.

### DIFF
--- a/Source/Flow/Private/Nodes/FlowNode.cpp
+++ b/Source/Flow/Private/Nodes/FlowNode.cpp
@@ -32,6 +32,7 @@ UFlowNode::UFlowNode(const FObjectInitializer& ObjectInitializer)
 #if WITH_EDITOR
 	Category = TEXT("Uncategorized");
 	NodeStyle = EFlowNodeStyle::Default;
+	bCanDelete = bCanDuplicate = true;
 #endif
 
 	InputPins = {DefaultInputPin};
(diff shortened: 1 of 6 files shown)


MothDoctor:
@ryanjon2040 Could you provide repro for this? Couldn't remove the Start node even prior to applying this change.

I also wonder, could you provide your examples of using methods like `CanDeleteNode` in node classes?
I assumed that would be limited to very specific cases like Start node, so didn't expose on `UFlowGraphNode` itself. I can't see anything against applying this change, just would like know your use cases :)

ryanjon2040:
@MothDoctor For the repro, (before applying this change) create a flow graph open it and delete the Start Node. It shouldn't get deleted. The aim of this PR is to simplify it and not to create additional classes. Since Flow Graph can be extended/modified by users to fit their needs, there can be instances where devs need to create additional event nodes (like Start) that shouldn't get deleted. An example is from my Sandbox plugin, where I've created additional event nodes that shouldn't be deleted by the dev. With this PR, they can simply set 2 booleans for deleting/duplicating in their custom nodes if desired and flow graph will take care of the rest.

Graph

PR #72, opened 2021-11-16
Timer node can be stopped if required.

ryanjon2040:
User can stop the timer early if required.
timer-stop


MothDoctor:
How is that different from Skip/Skipped pins? ;)

ryanjon2040:
lol you are correct. My bad. The skip name was giving me an impression like to skip step time or sorta something like that. 😄 

But don't you think `Stop` will be a better name? And maybe rename `In` to `Start`?

MothDoctor:
Hey, @ryanjon2040 ended up not accepting this change, but... entered a new feature request - so we could set Display Names for pins. This would provide non-destructive way to fix some confusing or outdated Pin Names :)

https://github.com/MothCocoon/FlowGraph/issues/89

PR #77, opened 2021-11-24
Clear previously selected nodes before new node creation

BorMor:
I find it quite annoying when I create new node and previous selected nodes stay selected, I have to unselect all nodes and select new node again to align it properly.


MothDoctor:

PR #78, opened 2021-11-30
Double clicking on blueprint node opens it.

ryanjon2040:
Similar to sub graph node, double clicking on custom flow node blueprint opens it.
Animation1

### DIFF
--- a/Source/FlowEditor/Private/Asset/FlowAssetEditor.cpp
+++ b/Source/FlowEditor/Private/Asset/FlowAssetEditor.cpp
@@ -914,6 +914,10 @@ void FFlowAssetEditor::OnNodeDoubleClicked(class UEdGraphNode* Node) const
 				}
 			}
 		}
+		else if (UObject* BlueprintAsset = FlowNode->GetClass()->ClassGeneratedBy)
+		{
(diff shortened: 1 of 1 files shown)


MothDoctor:
@ryanjon2040 The reason I didn't merge it yet, it's because "double-click" can be used to open "Primary Asset" (like Level Sequence on the `PlayLevelSequence` node) instead of node definition. And the latter case isn't a common one - my assumption is once you do node, you need an open node blueprint. There's a context option on the node just for that :)

image

It would confuse users that double-click sometimes open Node definition instead of a primary asset. Especially that open native Flow Node also requires using the `Jump to Node Definition` option.

I understand this would be convenient for debugging if we could jump directly to the exact node instance. Sadly, it conflicts with another use case, like jumping to a relevant Level Sequence asset.
We would need to solve the UX issue here. Not sure, maybe some project setting would help?

ryanjon2040:
Ah ok the `Jump to Node Definition` is new to me 😄.

MothDoctor:
@ryanjon2040 Accepted this PR as well.

Made 2 improvements
- implement my proposal from above, added `NodeDoubleClickTarget` enum to editor settings
- calling the already existing `UFlowGraphNode::JumpToDefinition` method, which is able to open IDE if a given node is defined in C++

image

Thanks for yet another contribution. This double-click thing really needed to have the alternate mode :)

PR #80, opened 2021-12-16
Added reverse play

seimei0083:
Added a boolean variable in the "Play Level Sequence" node to let user choose to play forward or backward.


MothDoctor:
Hi @seimei0083 I'd like to accept your change, but can't merge it directly as now conflict occurs (not your fault).
Would you be eager to update your PR, so I could merge it directly and you'd appear on the contributor list? :)

MothDoctor:
Hi @seimei0083 just added this option, based on your PR :)

PR #83, opened 2022-01-03
Some Changes

MYTNB:
1. Fixed FlowAssetInstance Select Bugs
2. Add FlowPrivateNode to Filter the SubFlowGraph
3. Add Graph Node Debug Info
4. Add Tick Event Define

https://user-images.githubusercontent.com/31533351/147934196-2196bc82-f68e-4d1f-8f19-4cfdab9c4b81.mp4

### DIFF
--- a/Source/Flow/Private/FlowAsset.cpp
+++ b/Source/Flow/Private/FlowAsset.cpp
@@ -3,6 +3,7 @@
 #include "FlowSubsystem.h"
 
 #include "Nodes/FlowNode.h"
+#include "Nodes/FlowPrivateNode.h"
 #include "Nodes/Route/FlowNode_CustomInput.h"
 #include "Nodes/Route/FlowNode_Start.h"
 #include "Nodes/Route/FlowNode_Finish.h"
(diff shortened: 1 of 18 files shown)


MothDoctor:
Hi @MYTNB 
would be possible for you to actually make separate PRs for every point. I'm sorry, it's not feasible to merge all of these as one changelist/commit even if I'd like to accept all of them.

I'd definitely accept a toolbar fix!
And displaying Error on the graph node :)

I'm not gonna accept adding Tick to the base `Flow Node` class. This would break a fundamental design of the graph, enforcing users to use event-based design instead of spamming Tick everywhere.
If Tick is needed in some specific node, it can be easily added by adding `FTickableGameObject`.

A proper explanation would be required for functional changes. Not sure what `FlowPrivateNode` introduces, and the class name itself is confusing.

The last thing, commit cannot include any non-Latin text as it would be a mystery for developers :)

MYTNB:
Thanks for your reply, I'm sorry for the trouble my commit caused you :).
FlowPrivateNode is designed to make the node  is only allowed to be used by some FlowAsset subtypes (Obviously I didn't find a suitable naming)
And I will not continue to request this submission (as I have already made very many changes to your base)

PR #86, opened 2022-01-25
Add input outputs to flow graph

ryanjon2040:
Alright here it is. This PR adds support for input and output pins 🥳. Now before you get excited lets talk about some limitations of this feature although these limitations can be lifted.

Screenshot_1

**FEATURES**
* Input output pins
* Supports per pin tooltip
* Supports any UObject/AActor (including custom ones too)
* Supports Float
* Supports Bool
* Supports Byte
* Supports Int
* Supports Int64
* Supports FGameplayTagContainer
* Supports FGameplayTag
* Supports FVector
* Supports FRotator
* Supports FTransform
* Supports FString
* Supports FText
* Supports FName
* Easy to set input or output

**LIMITATIONS:**
* Only one input/output is supported.
* Containers are not supported (Arrays, Set, Maps etc)

**UPCOMING** (If this gets accepted)
* Saving variables for later use (local and global)
* Calling graph event with input from Blueprint (wildcard pin)


MothDoctor:
Sure, there would a chance to accept it. 
What prevents it now
- Needs to overcome limitations, especially only single pin supported
- Moving property-specific code out of the Flow Node class, is possible. I mean part of code parsing specific types and handling value write/read.
- Need to compare with other methods like what Epic did in State Tree (UE5), but it might be less advanced than what we need for Flow Graph :)

ryanjon2040:
Yeah the single pin is kinda limiting but atleast the base is there so it can be worked on :)

Do you plan to merge this in its current state? Probably a separate branch or something so the community can use it for now? Or do you want to keep it open only?

Can you expand on the second part of your reply? Move out of Flow now as in a separate class, struct etc?

MothDoctor:
Well. practically this PR serves as a feature branch already? Any changes submitted to `ryanjon2040:input_output` should be automagically reflected here, right? :)
So unless it's in some way inconvenient for you, I would keep working on it here.

We could merge it even to the main branch, but after the 1.2 release - I'm planning a quick release to include all other PRs, remaining editor QoL features. It shouldn't take too long.

I was thinking about encapsulating operations on properties in a separate helper class, like engine's `PropertyPathHelpers` :)
image

ryanjon2040:
Ya any updates on my branch should reflect here. I'll work on the multi inputs once I can allocate some time. In the mean time if you plan to merge this in current state let me know 🙂

MothDoctor:
> As for the separate helper class, would it be necessary since the encapsulating operations are only called on this class. If required I can move it to `FlowHelpers` class which I already created with this PR.

Might do, especially if we would name the class `FlowPropertyHelpers` or something like that. Or maybe moved to `FlowPin` entirely? <thinking aloud> 
I just suspect that's gonna be a lot more property-supporting code while having support for multiple properties, containers, structs... ;)

ryanjon2040:
I've renamed `FlowHelpers` to `FlowPropertyHelpers` and put property operations under `SetPropertyValue` function. Commit 5b44d48cef0f057d2f3c810be667c0bf68ec8189

sturcotte06:
Maybe we should think about integrating UE5's `FInstancedStruct` instead of handling engine struct types directly. It costs an additionnal heap allocation, but it would allow to have any user-defined structs as input/output.

MothDoctor:
> Maybe we should think about integrating UE5's `FInstancedStruct` instead of handling engine struct types directly. It costs an additionnal heap allocation, but it would allow to have any user-defined structs as input/output.

Indeed! I'm thinking to research it, once I'll get a read of the standard graph QoL. A lot of work before even adding the next layer of data.

It's interesting especially given that Epic keeps improving. StateTree implementation got upgraded with an even newer `FInstancePropertyBag`, reducing boilerplate.
https://github.com/EpicGames/UnrealEngine/commit/af7394cf2731ae519b75d29aaffc63f7868aaa3f

It would obviously mean that variable support in this form would come in 5.1+ versions of the plugin. That wouldn't actually cause any delay, as I don't plan to implement it quickly. I experienced how much work it requires even if this is just in-house code. Open source is more challenging.

MothDoctor:
Hi, @ryanjon2040 finally decided not to pull this one in. 

- There are simply far too many limitations that haven't been addressed. I'm unable to finish this up, can't make it my responsibility at this point.
- UE5 `FInstancedStruct` and `StructUtils` in general. 5.1 develops further and State Tree already comes with an implementation sample.

image

Obviously, the community can still use your work, and pulled that in on their own. I do hope some people will find it useful, or they already did.

PR #87, opened 2022-01-26
Make UFlowNodeBlueprint not final to allow for subclassing

TheEnbyWitch:
Pretty straightforward change, which should make it so that you can create other types of UFlowNode blueprint assets in the game's editor module without having to edit Flow's source code.
image


MothDoctor:
That was insta-merge! Thank you! :)

PR #90, opened 2022-02-02
Added a custom dilation

seimei0083:
While playing the level sequence, get the custom dilation from the parent actor.


MothDoctor:
Accepted! Thanks for your contribution!

Applied some cosmetic changes. 

Functional changes
- And moved caching PlayRate from constructor (applied once to the node class) to `InitializeInstance` method that is called before executing a node.
- Exposed option to control if a user actually wants to apply CustomTimeDilation from the actor. Set to True by default.
- Added failsafe: in case Root Flow was created directly from some actor, not from The Flow Component.

Please check if everything work correctly for you :)

PR #92, opened 2022-02-16
Added DeniedClasses as option to blacklist FlowNodes

Mephiztopheles:
With the AllowedClasses we have already a white-list. With this blacklist, we want to add more convenience for the developers to allow all nodes by default but exlcude specific nodes


MothDoctor:
Accepted! Thanks for your contribution!

PR #102, opened 2022-05-10
Implemented the feature that can set the flow graph asset owner actor as the original start point for the level sequence.

seimei0083:


MothDoctor:
@seimei0083 Accepted this pull request :)

However, needed to make the following fixes.
- Removed unused `GraphOwner` property.
- Slightly renamed `bUseGraphOwnerAsOriginalPointActor` and commented on it, as it would be unclear to users what this does.
- Changed behavior: passing actual `TransformOriginActor` instead of just `FTransform`, which allows to dynamically read transform during playback. It might require updating your game-specific code, after merging plugin update.
- Added null check: setting LevelSequenceActor transform was reading `TransformOriginActor` without checking is it valid. Although updating this transform might be indeed useful in multiplayer games, for checking replication distance. Nice idea :)

Also added a simple example of how to use this feature pWhac-A-Mole stub in the FlowSolo project]{https://github.com/MothCocoon/FlowSolo/commit/ef853b45bd6e895a23e4fb1aa6a998648aced932).

PR #108, opened 2022-06-15
Fix add output pin + icon

dnault1:
Changes the FeditorStyle brush to the one used in the add input pin for the + icon


MothDoctor:
Thanks @dnault1!
Merged in and fixed also icon for inputs pins :)

PR #109, opened 2022-06-22
Added "Once" to OR

Mephiztopheles:
The OR-Operator can now be executed once and can be reset if, it's only executed once


MothDoctor:
Hey, @Mephiztopheles I wanted to accept your PR. Merged it, and worked a bit on improvements and it lead to discovering possible issues as the plugin's code needs some adjustments.

Sadly couldn't add Reset the pin at this time, as there could be issues with adding a new numbered pin. It's not your fault, these parts of the code weren't designed with adding regular inputs to numbered-input nodes. That's something to improve, added #113

Issue #16, opened 2021-01-31
Provide method to create nodes based on actor selected in the level


MothDoctor:
This way adding a reaction on the trigger would be as quick as with Level Blueprints.

MothDoctor:
Tried to implement that. It's actually more complex than it seems, design-wise and implementation-wise.
And it doesn't seem to make that big difference, especially since UE5 UI related to `TSoftObjectPtr<AActor>` is very convenient.

Dropping that ticket.

PR #110, opened 2022-06-25
Resolving SoftActorPtr using level from asset owner

Bargestt:
Just leaving an idea

Having quick actor ref in flow graph makes development super fast and straightforward


MothDoctor:
Idea looks good. I'll work on it together with #16 :)

MothDoctor:
Hey, @Bargestt I contemplated this matter a bit and decided that such a use case might be specific to a plugin I'm trying to keep minimal as possible.

If you'd still like to share this node with others, perhaps it would be better to set up a separate repository of various Flow Nodes? What do you think?

Bargestt:
Well, considering how many plugin variations were created by community it would be nice to create extension plugins
It will be pain to track and maintain though.

PR #112, opened 2022-08-13
Support for Lyra

scratali:
Remove the need for FlowWorldSettings so this plugin can be used in a Lyra based project


MothDoctor:
Honestly, I have no idea what happened here, as we have so many commits and changes here. It's like merging a version of a plugin from a year ago ;)

Also, I don't understand a change in a core element of a system to, allegedly, support a sample project. Which would wreak havoc on existing projects. It's not explained in the description.

PR #114, opened 2022-08-20
Rename uses of FEditorStyle to FAppStyle as it's been deprecated in 5.1 …

DoubleDeez:
…and add missing forward declarations and includes to fix unity compile issues

Main branch of unreal removed `FEditorStyle` stating that `FAppStyle` should be used instead and I was hitting a bunch of compile errors due to missing includes/forward declarations

### DIFF
--- a/Source/Flow/Public/FlowAsset.h
+++ b/Source/Flow/Public/FlowAsset.h
@@ -5,6 +5,7 @@
 #include "CoreMinimal.h"
 #include "FlowSave.h"
 #include "FlowTypes.h"
+#include "Templates/SubclassOf.h"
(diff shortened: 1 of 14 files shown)


MothDoctor:
@DoubleDeez I'm afraid this change doesn't compile on 5.0 ;)

If you'd like to update PR to just include non-unity fixes, I can merge it.
And 5.1 fixes I'd apply myself once UE 5.1 Preview is available :)

MothDoctor:
@DoubleDeez Your commit has been restored, as the plugin now has an actual 5.1 branch!
Thanks :)

PR #115, opened 2022-08-30
Sequence replication option

ArseniyZvezda:


MothDoctor:
@ArseniyZvezda change accepted, but do you think is it possible to get rid of expensive NetMulticast, method `RPC_InitializePlayer`?

Could be simply called on asset replication?
image

ArseniyZvezda:
@MothDoctor It was tested on dedicated server, single process and separate processes. 
I note that the control rig does not work correctly in a single process, but I think that this is a bug in the control rig. It works good with separate processes.
About Multicast, I will check

MothDoctor:
Hey, as I didn't get a reply yet, I decided to revisit this myself :)
Please, feel free to comment or fix if I broke something. I only have a chance to test it up in this small project.

My follow-up changes.
* Changed `LevelSequenceActor` spawn to deferred spawn, as it should be. It was overlooked prior to your change.
* Setting the `bReplicatePlayback` flag instead of calling `SetReplicatePlayback`, as this flag is read in `ALevelSequenceActor::PostInitializeComponents()`.
* Remove NetMulticast entirely. It worked in all cases in my demo, but I'd encourage you to perform more tests in a real project.
* Exposed `bAlwaysRelevant` as a parameter that can be set from the Flow Node creating this Sequence Player.
* And did better integration with a Transform Origin code.

PR #117, opened 2022-09-03
Asset category merge by name

Bargestt:
This will allow to reuse BuiltIn categories. eg Gameplay. 
And it'll show pop up "Require restart" when changing relevant settings

Merging is kinda random when it comes to other plugins with custom categories, but if other plugin loads first or has similar merger it will share menu category


MothDoctor:
@Bargestt thanks for this, I wouldn't think about such convenience!

Accepted, only added minor readability tweak.

PR #118, opened 2022-09-03
Class picker for new assets

Bargestt:
Just a helpful class selector when you have multiple asset classes like with DataAssets. 
It's fully optional and disabled by default

I'm quite sure most of people who have created quest or dialogue systems have implemented some form of this


MothDoctor:
@Bargestt that's a great QoL improvement! Also backward-compatible 👍 

Thanks for that, I'm quite used to adding custom AssetTypeActions and Factory classes for any custom class. But I bet it gonna be useful for many people :)

PR #119, opened 2022-09-05
Custom output finish

Cchnn:
The default subgraph output pin (`finish`) feels kind of limitative. Often times, I'd like to rename it for a more descriptive name or make another `finish` output pin, which would allow me to finish the subgraph at the same time as I trigger an output pin in a different context. Just like a flow node using `TriggerOutput`, it's up to the individual output pins to decide whether or not to finish the current flow graph in my opinion. What do you think?

### DIFF
--- a/Source/Flow/Private/Nodes/Route/FlowNode_CustomOutput.cpp
+++ b/Source/Flow/Private/Nodes/Route/FlowNode_CustomOutput.cpp
@@ -7,6 +7,7 @@
 
 UFlowNode_CustomOutput::UFlowNode_CustomOutput(const FObjectInitializer& ObjectInitializer)
 	: Super(ObjectInitializer)
+	, bFinish(false)
 {
 #if WITH_EDITOR
 	Category = TEXT("Route");
@@ -20,13 +21,13 @@ void UFlowNode_CustomOutput::ExecuteInput(const FName& PinName)
 {
 	if (!EventName.IsNone() && GetFlowAsset()->GetCustomOutputs().Contains(EventName) && GetFlowAsset()->GetNodeOwningThisAssetInstance())
 	{
-		GetFlowAsset()->TriggerCustomOutput(EventName);
+		GetFlowAsset()->TriggerCustomOutput(EventName, bFinish);
 	}
 }
 
 #if WITH_EDITOR
 FString UFlowNode_CustomOutput::GetNodeDescription() const
 {
-	return EventName.ToString();
+	return EventName.ToString() + LINE_TERMINATOR + (bFinish ? TEXT("True") : TEXT("False"));
 }
 #endif
--- a/Source/Flow/Private/Nodes/Route/FlowNode_CustomOutput.cpp
+++ b/Source/Flow/Private/Nodes/Route/FlowNode_CustomOutput.cpp
@@ -7,7 +7,7 @@
 
 UFlowNode_CustomOutput::UFlowNode_CustomOutput(const FObjectInitializer& ObjectInitializer)
 	: Super(ObjectInitializer)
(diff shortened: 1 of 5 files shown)


Source/Flow/Public/Nodes/Route/FlowNode_CustomOutput.h
```diff
@@ -16,6 +16,9 @@ class FLOW_API UFlowNode_CustomOutput final : public UFlowNode
 
 	UPROPERTY()
 	FName EventName;
+
+	UPROPERTY(EditAnywhere, Category = "Output")
+	bool bFinish;
```
MothDoctor, line 21: A new property in this form is currently a mystery for a user. It needs a more descriptive name (`bFinishGraph` perhaps?) and comment above it. This way a random user could use it right away :)

Source/Flow/Private/Nodes/Route/FlowNode_CustomOutput.cpp
```diff
@@ -20,13 +21,13 @@ void UFlowNode_CustomOutput::ExecuteInput(const FName& PinName)
...
-		GetFlowAsset()->TriggerCustomOutput(EventName);
+		GetFlowAsset()->TriggerCustomOutput(EventName, bFinish);
 	}
 }
 
 #if WITH_EDITOR
 FString UFlowNode_CustomOutput::GetNodeDescription() const
 {
-	return EventName.ToString();
+	return EventName.ToString() + LINE_TERMINATOR + (bFinish ? TEXT("True") : TEXT("False"));
```
MothDoctor: "True"/"False" doesn't communicate what's happening. I would rather add the text "Finish Graph" or "Finish Graph = True" only if the property is set to True. This way it would stand out and communicate behavior.

Source/Flow/Private/Nodes/Route/FlowNode_SubGraph.cpp
```diff
@@ -18,7 +18,7 @@ UFlowNode_SubGraph::UFlowNode_SubGraph(const FObjectInitializer& ObjectInitializ
 #endif
 
 	InputPins = {StartPin};
-	OutputPins = {FinishPin};
+	OutputPins = {};
```
MothDoctor, line 21: SubGraph node needs to keep `FinishPin` output. Otherwise, we would break existing content and usage patterns. Finishing my custom event should be still optional.

MothDoctor:
Hi @Cchnn it's a good idea :)

I could accept it, but it needs a few tweaks. All are listed as review comments.

Source/Flow/Private/Nodes/Route/FlowNode_SubGraph.cpp
MothDoctor: fixed it post-merge

Source/Flow/Private/Nodes/Route/FlowNode_CustomOutput.cpp
MothDoctor: fixed it post-merge

MothDoctor (commented):
all done

MothDoctor:
OK, I've tried to integrate it. I was already in the progress of fixing it (the new option was only finishing the node, not the entire graph), but using it this way quickly appeared convoluted, and unclear to the user.

It might be not trivial to change the execution order - first trigger custom events, then finish the graph.

In the end, I don't see that chaining events this way is big trouble for regular users. So reversed this PR merge. It was worth the shot, though :)

image

However, I added the `CanFinishGraph` method to `UFlowNode`. It should make it easier to implement such custom behavior in a forked plugin :)

PR #120, opened 2022-09-05
On node input triggered/output triggered

Cchnn:
Added an `Activate` function. Currently, there's no proper `Initialize` function. Although we do have `InitializeInstance`, it only gets called once, unlike `Cleanup`. Let's say I initialize an object in `InitializeInstance` (`Object = NewObject()`) and I clean it up in `Cleanup` (`Object = nullptr`). I might still end up with `nullptr` if the flow node has already been cleaned up and I tried to trigger the same flow node again. I've noticed that some of the native flow nodes initialize its member variable in `ExecuteInput`. But I feel like having a proper `Initialize` function would be helpful, just like `BeginPlay/EndPlay` in `AActor`. What do you think?

`OnNodeInputTriggered` and `OnNodeOutputTriggered` allow us to do something (e.g. logging) whenever a flow node has its input/output pin triggered.

### DIFF
--- a/Source/Flow/Private/FlowSubsystem.cpp
+++ b/Source/Flow/Private/FlowSubsystem.cpp
@@ -48,6 +48,14 @@ void UFlowSubsystem::Deinitialize()
 	AbortActiveFlows();
 }
 
+void UFlowSubsystem::OnNodeInputTriggered(const UFlowNode* node, const bool bWasActive)
+{
+}
+
+void UFlowSubsystem::OnNodeOutputTriggered(const UFlowNode* node, const bool bFinish)
+{
+}
+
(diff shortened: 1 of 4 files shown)


MothDoctor:
Hi @Cchnn, I'm afraid I cannot accept the change to Flow Subsystem, it shouldn't ever know about the execution of pins on specific nodes. It would break encapsulation.

I could accept the commit only adding an `Activate` event. I don't see a need for need it personally, but I guess other people would find it convenient to use :)

MothDoctor:
OK, I did accept your change, just to simply cut everything except `Activate` event. Thanks for your contribution! :)

PR #122, opened 2022-09-26
Hide advanced pins for specific flow node classes

Cchnn:
It could come in handy if you have a flow node with a lot of pins. What do you think?


MothDoctor:
Mhm, it looked fine initially. But... after consideration, I'm decided not to merge it.

The reason is that might severely break graphs (as pins with connections would be hidden), and it might also impair the debugger. In effect, it might confuse users and introduce new bugs to the system.

PR #124, opened 2022-11-15
Refresh Node - Fix mapping existing nodes to their new classes

ArseniyZvezda:
I posted this method on discord 08/30/2022. It also helped a few people besides me. It may need a little polishing, but it does what it's supposed to.
Tested on 5.0

### DIFF
--- a/Source/FlowEditor/Private/Asset/FlowAssetEditor.cpp
+++ b/Source/FlowEditor/Private/Asset/FlowAssetEditor.cpp
@@ -33,6 +33,7 @@
 #include "ScopedTransaction.h"
 #include "SNodePanel.h"
 #include "ToolMenus.h"
+#include "Nodes/Route/FlowNode_Start.h"
 #include "Widgets/Docking/SDockTab.h"
 
 #define LOCTEXT_NAMESPACE "FlowAssetEditor"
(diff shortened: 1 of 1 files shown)


MothDoctor:
Yes, you're right about code being in need of polish ;)
I wanted to finish up myself, but this change became too big. Especially today this a perfect day to release a new plugin version, because of the 5.1 release. Reverting changes then, we'll revisit after release :)

Or you could try to update to create a brand new PR based on this feedback. (Unless there's some magical way to re-open this one).
Either way, it's definitely worth implementing this.

Review
* First of all, such fixup code is better in `UFlowGraph::PostLoad` than `FFlowAssetEditor::RefreshAsset` as it now graphs can be fixed upon load. It shouldn't be a manual fixup. `UFlowGraphNode` would be a better place, if we could ensure that the node of the given class always exists, but I understand that core issue here, so... I would propose trying to patch it up via `UFlowGraph::PostLoad`.
* Especially that you also happened to delete the entire `RefreshContextPins` functionality ;)
* Code for creating nodes should reuse `FFlowGraphSchemaAction_NewNode::CreateNode`, we should extend this method. So it can receive instances of the old Graph Node, and pass its data to the new instance. Or we could add a very similar class to the same file.
* In my trial, I added searching old Graph Nodes by FGuid, if needed. That replaces iterating all Graph Nodes every single time (first loop in your implementation), even if all nodes are fine.

MothDoctor:
@ArseniyZvezda rework done and submitted :)

EDIT And... returned to calling it manually from `Refresh Asset`, as the automatic PostLoad fixup had some issues... :(

PR #126, opened 2022-11-18
Use in-built SetEnabledState instead of overriding non-virtual methods

ryanjon2040:
This changes gives you the same result as before. Benefit is we don't have to override the following methods. Basically less clutter.

**GetNodeTitleColor
GetNodeBodyColor
GetNodeTitleIconColor
GetNodeTitleTextColor**


MothDoctor:
Thanks for the suggestion, but I omitted the built-in UEdGraphNode state which is used for blueprints. It may change at any time and may be confusing for users. We don't use blueprint enabled/disabled state in execution, so it is misleading to read it in UI.
Plus it provides a way to customize node appearance. I wouldn't like to be limited here by a blueprint look'n'feel :)

I know that I'm replacing non-virtual methods. I consider this a thing that should be changed by

PR #130, opened 2023-02-05
Allow Sequence nodes to execute new connections

jhartikainen:
This is an update to the Sequence node, based on what we discussed shortly on Discord.

This adds a new flag to Sequence, which makes it track which connections it executed previously. If new connections are added or old ones are changed, it can then execute the changed connections when loading a save game.

Reasoning: This allows the developer to add new logic into existing flowgraphs, without causing issues in savegames. Without this feature, if a player has saved the game and new logic is added into a graph, there is no mechanism to ensure the logic is ran when the game is loaded. With this change, the developer can simply update the sequence connections and new nodes will execute after the game is loaded.

I don't know if this is the optimal/best way to go about it, but I figured I'll open this PR for any ideas/discussion since I had this code anyway.

# Example scenarios:

## Add a new connection:

- Developer creates a sequence and enables the save pin execution flag.
- Developer adds one connection to the node
- Player plays, the sequence is executed, player saves the game
- Developer adds a new connection to the sequence node
- Players load game, since the save flag was enabled, the newly added connection is executed, allowing the player to get the new features

## Replace a connection

Same as above, but instead of adding a new pin, the developer can replace an existing connection with a new node.
The code stores the FGuid of the connected node, so it can detect when the connected node changes.

### DIFF
--- a/Source/Flow/Private/Nodes/Route/FlowNode_ExecutionSequence.cpp
+++ b/Source/Flow/Private/Nodes/Route/FlowNode_ExecutionSequence.cpp
@@ -16,10 +16,48 @@ UFlowNode_ExecutionSequence::UFlowNode_ExecutionSequence(const FObjectInitialize
 
 void UFlowNode_ExecutionSequence::ExecuteInput(const FName& PinName)
 {
+	if(bSavePinExecutionState)
+	{
+		ExecuteNewConnections();
+		return;
+	}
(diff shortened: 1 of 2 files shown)


MothDoctor:
Submitted change to the Sequence node: bSavePinExecutionState is now True by default. Thanks again! :)

PR #132, opened 2023-02-12
Add null flow asset check when calling StartRootFlow

bohdon:
Hello! First of all, amazing plugin and thanks so much for making it open source 🙏🏼 

My first time testing things out I accidentally just called FlowSubsystem->StartRootFlow in blueprint with no flow asset and it crashed. This fix adds a simple null check and uses the message log to surface the error to the designer. I know you haven't used FMessageLog much in the plugin so far but I find it extremely helpful to get an in-your-face error for blueprint things like this. I also decided to put the null check directly in the StartRootFlow function since it is the blueprint callable function. Calling CreateRootFlow with a null asset will still crash, but you've got the check in CreateFlowInstance and it's native only so a programmer should know what's up.

Cheers!


MothDoctor:
Change accepted, thanks! :)
Applied cosmetic readability changes, and fixed non-editor compilation.

PR #133, opened 2023-02-12
add bExactMatch option to Notify Actor flow node

bohdon:
Hello, this change adds a `bExactMatch` option to the Notify Actor node. `FlowSubsystem->GetComponents` already has that built in functionality, and adding this option allows out of the box broadcasting to components in a more versatile way, e.g. 'disable all interactions'. I realize the performance concern, and kept that note in the property description, but for non-open-world games this seems like a nice core built in feature so that designers don't have to add all exact identity tags to a flow component: e.g. `Object` and `Object.Interaction` and `Object.Interaction.Store`, just the last one.

Thanks!


MothDoctor:
PR accepted, thanks! :)

PR #134, opened 2023-02-13
Add "delay until next tick" node

ryanjon2040:
Similar to the Blueprint node, I added one to Flow Graph.
image


MothDoctor:
I wonder if we could add this as an option to the Timer node. This is how it also works if set the Delay time to 0.0 - it will fire output on the next tick. You're even using here method called `SetTimerForNextTick`.

I'm a bit resistant to adding such rarely used nodes like this to core nodes.

ryanjon2040:
Sure thing. I'll add this to the Timer node so if it is 0, it will fire off in next tick. Feel free to close this PR.

PR #135, opened 2023-02-13
Add custom color support to nodes

ryanjon2040:
Setting Node Style to custom will let the user colorize the node to his/her own color. 

NOTE: This is prioritized over **UFlowGraphSettings->NodeSpecificColors** and **UFlowGraphSettings->NodeTitleColors** which means if you set to Custom then _NodeSpecificColors_ and _NodeTitleColors_ are ignored for that node.

NodeColor


MothDoctor:
That's a great idea, accepted! Thanks :)

ryanjon2040:
Thank you so much for accepting this. 🙂

PR #137, opened 2023-02-15
Add keywords for timer to make search easier

ryanjon2040:
A small change that adds a few useful keywords to Timer node


MothDoctor:
Merged thanks! :)

MothDoctor:
PS Only removed "wait" from keywords as many game-specific nodes can have names like "wait on spawn". Finding Timer on list here might be confusing.

ryanjon2040:
That's alright. Glad this got merged :)

PR #138, opened 2023-02-16
Add option to loop indefinitely in Timer node

ryanjon2040:
New option to loop the timer indefinitely. If the option is true, it will execute **Step** pin every **_StepTime_** unless **Skip** is called.


MothDoctor:
Hi, finally I decided to skip this PR. It's nothing technically wrong with the code itself. You see... it would allow inexperienced designers to abuse infinite Timer like they would abuse blueprint Tick, by using it instead of proper event-based architecture enclosed in specific nodes.

It might be a better idea to make a separate "tick" node to support cases where we really want to repeat some actions indefinitely. This way we would also avoid making a single node too complicated :)

PR #140, opened 2023-02-20
Show pretty readable pin name even if friendly name is not provided

ryanjon2040:
This change shows friendly name of the pin even if **_PinFriendlyName_** in **FFlowPin** is not provided so users don't have to keep setting readable name in **_PinFriendlyName_**. Of course they can still override and set it if they wish.

image

This behavior is toggleable from **Editor Preferences** -> **Appearance** -> **Show Friendly Variables Names**
image


MothDoctor:
Finally accepted this, thanks! Two modifications were needed.

1. added `bEnforceFriendlyPinNames` to the plugin
- this allows controlling Flow Graph behavior separately from the properties and blueprint functions
- engine display name of that engine's flag is already confusing, they should split it in two ;)

2. Moved creation of display pin name to ` UFlowGraphSchema::GetPinDisplayName`. Code in Flow Graph Node was actually baking strings into node instance. That would require recreating nodes in order to apply the `bEnforceFriendlyPinNames` flag which is supposed to be a user-specific setting.

PR #141, opened 2023-02-20
Properly deprecate LogError

ryanjon2040:
Previously it was only deprecated for Blueprint. Now it is for C++ too.


MothDoctor:
Oh, awesome, thanks again :)

PR #142, opened 2023-02-20
Timer node complete in next tick if value is closer to 0

ryanjon2040:
As mentioned in #134 I've made the Timer node complete in next tick if value is 0. This PR might conflict with #138.


MothDoctor:
In this case, I would use be in favor of creating a separate node like "next tick", as this gets a bit confusing to test timer value against `UE_KINDA_SMALL_NUMBER`.

And again I'm afraid of adding this to the plugin, as inexperienced designers would abuse it. They might use it instead of learning event-based architecture as they do with "Delay = 0.0 seconds" in blueprints.
IMHO such nodes would great addition to hypothetical "community nodes" repository :)

ryanjon2040:
> In this case, I would use be in favor of creating a separate node like "next tick", as this gets a bit confusing to test timer value against `UE_KINDA_SMALL_NUMBER`.

That "next tick" node is what I did in #134 but there you suggested to add this functionality to Timer node.

MothDoctor:
> > In this case, I would use be in favor of creating a separate node like "next tick", as this gets a bit confusing to test timer value against `UE_KINDA_SMALL_NUMBER`.
> 
> That "next tick" node is what I did in #134 but there you suggested to add this functionality to Timer node.

Oh, right. I apologize for that, I'm already in vacation mode. I'll pull it after I'm back home.
Could you fix merge conflicts, please? :)

ryanjon2040:
> > > In this case, I would use be in favor of creating a separate node like "next tick", as this gets a bit confusing to test timer value against `UE_KINDA_SMALL_NUMBER`.
> > 
> > 
> > That "next tick" node is what I did in #134 but there you suggested to add this functionality to Timer node.
> 
> Oh, right. I apologize for that, I'm already in vacation mode. I'll pull it after I'm back home. Could you fix merge conflicts, please? :)

Are you planning to pull this commit or #134? Because you said to fix merge conflicts, but #134 doesn't have any.

MothDoctor:
@ryanjon2040 I resolved conflicts in this commit via web and accepted this change! Thanks again! :)

PR #143, opened 2023-02-21
Allow signal mode to be set at runtime

ryanjon2040:
With this new node (and callable from C++) you can change signal mode at runtime. I made this for my use case where I need to run certain nodes only once and should technically disable itself once it runs its course. So instead of adding more logic specific to my needs I thought I'll make use of the existing signal mode. Setting **SignalMode** directly in Flow Node doesn't update it visually and once PIE ends it won't change back to the original state either. So I modified to make sure when this function is called, it will visually change the node (see the gif) and will reset back to the original state when PIE ends.

image

SetSignalMode

### DIFF
--- a/Source/Flow/Private/FlowAsset.cpp
+++ b/Source/Flow/Private/FlowAsset.cpp
@@ -82,6 +82,22 @@ EDataValidationResult UFlowAsset::ValidateAsset(FFlowMessageLog& MessageLog)
 	return MessageLog.Messages.Num() > 0 ? EDataValidationResult::Invalid : EDataValidationResult::Valid;
 }
 
+void UFlowAsset::ResumePIE(const bool bIsSimulating)
+{
+	OnResumePIE(bIsSimulating);
+}
+
+void UFlowAsset::EndPIE(const bool bIsSimulating)
+{
+	// Intentionally using RecordedNodes. We only need to reset signal mode if the node was ever called. @See SetSignalModeRuntime in FlowNode
+	for (UFlowNode* Node : RecordedNodes)
+	{
+		Node->ResetSignalMode();
+	}
+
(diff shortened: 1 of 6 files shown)


Mephiztopheles:
Can you give an example for a use case?

ryanjon2040:
> Can you give an example for a use case?

@Mephiztopheles For my RTS game I'm using Flow Graph to script the AI behavior and I have 3 primary nodes (where all the child nodes are inherited) and one of them I call is Rule nodes. Rule nodes executes if certain conditions are met and some rule nodes needs to disable itself once they are success. For example, AI can check if certain conditions are met to construct a building and if it is a success, construct the building and disable the rule node to stop processing it further.

Here is an example graph. The timer runs every 0.75 seconds. The disable pin will set the signal mode to Pass Through once the **Construct Building** node is success.
image

Mephiztopheles:
Ok, so with the variable-support, you could just use if-nodes, right?
It's basically do once per graph

ryanjon2040:
> Ok, so with the variable-support, you could just use if-nodes, right?
> It's basically do once per graph

@Mephiztopheles Actually with #136 I could've skipped this PR. That one opens a lot more possibilities.

MothDoctor:
As pointed out above, flow control can be - and should be - handled visibly in the graph. This is a fundamental design assumption here. From this perspective, changing signal mode in runtime would be hacky. It wouldn't be clear why some nodes stopped working. 
Signal mode is meant to deactivate nodes in the game patch :)

Many projects implemented their own `Facts DB`, which allow routing signal in graph based on any previous action.
And we can have basic variables support soon thanks to your work :)

ryanjon2040:
Thank you @MothDoctor. I'll handle it manually as you said within my nodes :)

PR #144, opened 2023-02-24
More keywords to make search easier for multiple nodes

ryanjon2040:
Added different keywords to multiple nodes which I think should help with search.


MothDoctor:
Accepted!
Made a few tweaks to avoid confusion, i.e. many projects have separate "cutscene" or "movie" nodes. It would be confusing if basic `PlayLevelSequence` node would show up in the search with custom nodes.
https://github.com/MothCocoon/FlowGraph/commit/730c2b4de6ab33b2c6ff8e0d09b98cf136dab427

Thanks again! :)

PR #145, opened 2023-02-27
Add an option to toggle signal mode logs

ryanjon2040:
For my use where I have a Timer that runs continuously and setting signal mode at runtime causes too much spam on the output log. By default these are enabled so it will work like before.

image


MothDoctor:
Accepted! Thanks again :)

PR #146, opened 2023-02-28
Add an option to toggle invoking of runtime tab

ryanjon2040:
New option to toggle automatic invoking of runtime message tab. To understand why this PR is a good idea to merge is to create the following graph and run it.

image

Now try to switch to another Blueprint or do anything in Unreal for the next 60 seconds 😜. Obviously the above graph is just an example but in any case, if someone wants to set a node pass through (or disabled) for the time being and experiment with something (while they execute that pass-through or disabled node in a timer or something), it can be an annoying experience being unable to do anything because of the invoking tab.

@MothDoctor I've set the option to false by default. In case you decide to merge this in, feel free to change `bAutoFocusOnRuntimeMessageAdded = true` if you want to.

### DIFF
--- a/Source/FlowEditor/Private/Asset/FlowAssetEditor.cpp
+++ b/Source/FlowEditor/Private/Asset/FlowAssetEditor.cpp
@@ -39,6 +39,7 @@
 #include "ScopedTransaction.h"
 #include "SNodePanel.h"
 #include "ToolMenus.h"
+#include "Graph/FlowGraphSettings.h"
 #include "Widgets/Docking/SDockTab.h"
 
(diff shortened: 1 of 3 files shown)


MothDoctor:
Mhm, I tried that repo. Flow Graph only focused once... Anyway, this autofocusing Runtime Log doesn't seem to be a good idea in general. 

So I've started work on some refactoring inspired by your report :)
- moved handling of the runtime log to the editor subsystem
- editor subsystem throws notifications to users on End PIE "hey, this asset has an issue, click to open asset editor"

I'll submit it soon, so it doesn't make sense to accept this PR directly. After refactor editor won't try to call autofocus at all.
Thanks for a great report!

PR #147, opened 2023-03-01
Add an option to hide node description when PIE is active

ryanjon2040:
For quick readability you can hide node description temporarily while PIE is active. By default it is off so old behavior is kept.

GIF:
NodeDescriptionHide


MothDoctor:
Accepted, thanks! :)

I only changed the config property name to `bShowNodeDescriptionInPIE` as double negation is harder to understand by human brain ;)

MothDoctor:
FYI renamed the flag to `bShowNodeDescriptionWhilePlaying` as it doesn't only impact PIE.

PR #151, opened 2023-03-20
Extending the "GetAssignedGraphNodeClass()" function to support to return the correct "EdGraphNode" for grand child classes of "UFlowNode"

Vi-So:
- Extending the "GetAssignedGraphNodeClass()" function to support to return the correc "EdGraphNode" for grand child classes of "UFlowNode" -

Previously it would return the first found "IsChildOf" and return the EdGraphNode of the foundclass.

But lets say we got this situation just for example: 
UFlowNode->UMyBaseNode
UFlowNode->UMyBaseNode->UMyQuestNode

Both "UMyBaseNode" and "UMyQuestNode" could be in the list if we want different EdGraphNodes for both of them. If the "UMyQuestNode" goes into the itteration and it would find "UMyBaseNode" first it will be returned without checking the rest, because "UMyQuestNode" is dirived from "UMyBaseNode" (IsChildOf == true) and thus it returns the EdGraphNode of the base node directly. So that would mean I got the wrong EdGraphNode for my special quest node.

This modification continues to try and find parent classes and returns the closest parent version. 
So lets say we have this situation:
UFlowNode->UMyBaseNode
UFlowNode->UMyBaseNode->UMyQuestNode->MyOtherQuestNode(BP class) 

Both “UMyBaseNode” and “UMyQuestNode” would be found, but the closest parent to “MyOtherQuestNode” would be returned, in this case the EdGraphNode of “UMyQuestNode”

It only does this if we have found 2 or more parents though. If only 1 is found we would return that one, and if the exact class match is found lets say: "UMyQuestNode" finds "UMyQuestNode" it will return the EdGraphNode right away aswell.

Hopefully this makes sence :P Otherwise feel free to ask.

### DIFF
--- a/Source/FlowEditor/Private/Graph/FlowGraphSchema.cpp
+++ b/Source/FlowEditor/Private/Graph/FlowGraphSchema.cpp
@@ -237,15 +237,54 @@ TArray<TSharedPtr<FString>> UFlowGraphSchema::GetFlowNodeCategories()
 
 UClass* UFlowGraphSchema::GetAssignedGraphNodeClass(const UClass* FlowNodeClass)
 {
+	TArray<UClass*> FoundParentClasses;
+	UClass* ReturnClass = nullptr;
+
+	// Collect all possible parents and their corresponding GraphNodeClasses
 	for (const TPair<UClass*, UClass*>& GraphNodeByFlowNode : GraphNodesByFlowNodes)
 	{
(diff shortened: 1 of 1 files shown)


MothDoctor:
Awesome, I just stumbled on a similar problem but didn't resolve it on the plugin level. Thanks!
PS Only made a single cosmetic tweak to your code.

PR #153, opened 2023-04-26
Fixed bug where Owner parameter wasn't being used 

LindyHopperGT:
Owner was passed into this UFlowComponent::GetRootInstances(), presumably to be used instead of 'this', but it was just using 'this'.

Changed the code to use Owner if specified, and default to 'this' otherwise.


MothDoctor:
Merged, thanks! :)

PR #154, opened 2023-05-01
CustomInput & Output Details Customization and other cleanup

LindyHopperGT:
**Simple changes:**
* Added forward declarations & includes that were needed to compile
* Changed some functions to return const &'s rather than doing a full deep copy of a member container
* Explicitly calling MoveTemp() to use move semantics (supported by UE container classes (eg TArray)) to move the array without an extra copy.  Some compilers might be able to do this optimization automatically, but being explicit here.
* Exposed CustomInput Add/Remove functions on UFlowNode to allow subclasses to modify the CustomInputs array.

**More interesting changes:**
* Introduced a superclass to UFlowNode_CustomInput and Output so they can share more code.  The diff is easier if you compare them to the new superclass, since most of the code moved into it.
* The "details" superclass for CustomInput/Output fixes a discovered bug where the list of EventNames was not recaching correctly.
* Moved a bit of code from PlayLevelSequence to FlowNode.h (TryGetRootFlowActorOwner) so that it can be used in other contexts.  Also provided a component version of this same code.
* The FlowGraphSchema will now create default nodes for any CustomInputs that exist when the asset is first created.
* Added a UseAdaptiveNodeTitles option to optionally make CustomInputs integrate their EventName into the title for the node.  This defaults to false (to preserve previous behavior by default).

### DIFF
--- a/Source/Flow/Private/FlowAsset.cpp
+++ b/Source/Flow/Private/FlowAsset.cpp
@@ -190,6 +190,8 @@ void UFlowAsset::HarvestNodeConnections()
 	}
 }
 
+#endif // WITH_EDITOR
+
 void UFlowAsset::AddCustomInput(const FName& InName)
 {
 	check(!CustomInputs.Contains(InName));
@@ -203,8 +205,6 @@ void UFlowAsset::RemoveCustomInput(const FName& InName)
 
 	CustomInputs.Remove(InName);
 }
-
-#endif // WITH_EDITOR
 
 UFlowNode_Start* UFlowAsset::GetStartNode() const
 {
--- a/Source/Flow/Private/FlowAsset.cpp
+++ b/Source/Flow/Private/FlowAsset.cpp
@@ -189,7 +189,22 @@ void UFlowAsset::HarvestNodeConnections()
 		}
 	}
 }
-#endif
+
+void UFlowAsset::AddCustomInput(const FName& InName)
+{
+	check(!CustomInputs.Contains(InName));
+
+	CustomInputs.Add(InName);
+}
+
+void UFlowAsset::RemoveCustomInput(const FName& InName)
+{
+	check(CustomInputs.Contains(InName));
+
+	CustomInputs.Remove(InName);
+}
+
+#endif // WITH_EDITOR
 
 UFlowNode_Start* UFlowAsset::GetStartNode() const
 {
--- a/Source/Flow/Private/FlowSubsystem.cpp
+++ b/Source/Flow/Private/FlowSubsystem.cpp
@@ -11,6 +11,7 @@
 
 #include "Engine/GameInstance.h"
 #include "Engine/World.h"
+#include "Logging/MessageLog.h"
 #include "Misc/Paths.h"
 #include "UObject/UObjectHash.h"
 
--- a/Source/Flow/Private/FlowSubsystem.cpp
+++ b/Source/Flow/Private/FlowSubsystem.cpp
@@ -265,7 +265,7 @@ TMap<UObject*, UFlowAsset*> UFlowSubsystem::GetRootInstances() const
 	{
 		Result.Emplace(RootInstance.Value.Get(), RootInstance.Key);
 	}
-	return MoveTemp(Result);
+	return Result;
 }
 
 TSet<UFlowAsset*> UFlowSubsystem::GetRootInstancesByOwner(const UObject* Owner) const
@@ -278,7 +278,7 @@ TSet<UFlowAsset*> UFlowSubsystem::GetRootInstancesByOwner(const UObject* Owner) 
 			Result.Emplace(RootInstance.Key);
 		}
 	}
-	return MoveTemp(Result);
+	return Result;
 }
 
 UFlowAsset* UFlowSubsystem::GetRootFlow(const UObject* Owner) const
@@ -537,7 +537,7 @@ TSet<UFlowComponent*> UFlowSubsystem::GetFlowComponentsByTags(const FGameplayTag
 		}
 	}
 
-	return MoveTemp(Result);
+	return Result;
 }
 
 TSet<AActor*> UFlowSubsystem::GetFlowActorsByTag(const FGameplayTag Tag, const TSubclassOf<AActor> ActorClass, const bool bExactMatch) const
@@ -554,7 +554,7 @@ TSet<AActor*> UFlowSubsystem::GetFlowActorsByTag(const FGameplayTag Tag, const T
 		}
 	}
 
-	return MoveTemp(Result);
+	return Result;
 }
 
 TSet<AActor*> UFlowSubsystem::GetFlowActorsByTags(const FGameplayTagContainer Tags, const EGameplayContainerMatchType MatchType, const TSubclassOf<AActor> ActorClass, const bool bExactMatch) const
@@ -571,7 +571,7 @@ TSet<AActor*> UFlowSubsystem::GetFlowActorsByTags(const FGameplayTagContainer Ta
 		}
 	}
 
-	return MoveTemp(Result);
+	return Result;
 }
 
 TMap<AActor*, UFlowComponent*> UFlowSubsystem::GetFlowActorsAndComponentsByTag(const FGameplayTag Tag, const TSubclassOf<AActor> ActorClass, const bool bExactMatch) const
@@ -588,7 +588,7 @@ TMap<AActor*, UFlowComponent*> UFlowSubsystem::GetFlowActorsAndComponentsByTag(c
 		}
 	}
 
-	return MoveTemp(Result);
+	return Result;
 }
 
 TMap<AActor*, UFlowComponent*> UFlowSubsystem::GetFlowActorsAndComponentsByTags(const FGameplayTagContainer Tags, const EGameplayContainerMatchType MatchType, const TSubclassOf<AActor> ActorClass, const bool bExactMatch) const
@@ -605,7 +605,7 @@ TMap<AActor*, UFlowComponent*> UFlowSubsystem::GetFlowActorsAndComponentsByTags(
 		}
 	}
 
-	return MoveTemp(Result);
+	return Result;
 }
 
 void UFlowSubsystem::FindComponents(const FGameplayTag& Tag, const bool bExactMatch, TArray<TWeakObjectPtr<UFlowComponent>>& OutComponents) const
--- a/Source/Flow/Private/FlowSubsystem.cpp
+++ b/Source/Flow/Private/FlowSubsystem.cpp
@@ -265,7 +265,7 @@ TMap<UObject*, UFlowAsset*> UFlowSubsystem::GetRootInstances() const
 	{
 		Result.Emplace(RootInstance.Value.Get(), RootInstance.Key);
 	}
-	return Result;
+	return MoveTemp(Result);
 }
 
 TSet<UFlowAsset*> UFlowSubsystem::GetRootInstancesByOwner(const UObject* Owner) const
@@ -278,7 +278,7 @@ TSet<UFlowAsset*> UFlowSubsystem::GetRootInstancesByOwner(const UObject* Owner) 
 			Result.Emplace(RootInstance.Key);
 		}
 	}
-	return Result;
+	return MoveTemp(Result);
 }
 
 UFlowAsset* UFlowSubsystem::GetRootFlow(const UObject* Owner) const
@@ -537,7 +537,7 @@ TSet<UFlowComponent*> UFlowSubsystem::GetFlowComponentsByTags(const FGameplayTag
 		}
 	}
 
-	return Result;
+	return MoveTemp(Result);
 }
 
 TSet<AActor*> UFlowSubsystem::GetFlowActorsByTag(const FGameplayTag Tag, const TSubclassOf<AActor> ActorClass, const bool bExactMatch) const
@@ -554,7 +554,7 @@ TSet<AActor*> UFlowSubsystem::GetFlowActorsByTag(const FGameplayTag Tag, const T
 		}
 	}
 
-	return Result;
+	return MoveTemp(Result);
 }
 
 TSet<AActor*> UFlowSubsystem::GetFlowActorsByTags(const FGameplayTagContainer Tags, const EGameplayContainerMatchType MatchType, const TSubclassOf<AActor> ActorClass, const bool bExactMatch) const
@@ -571,7 +571,7 @@ TSet<AActor*> UFlowSubsystem::GetFlowActorsByTags(const FGameplayTagContainer Ta
 		}
 	}
 
-	return Result;
+	return MoveTemp(Result);
 }
 
 TMap<AActor*, UFlowComponent*> UFlowSubsystem::GetFlowActorsAndComponentsByTag(const FGameplayTag Tag, const TSubclassOf<AActor> ActorClass, const bool bExactMatch) const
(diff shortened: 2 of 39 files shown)


Source/Flow/Private/FlowAsset.cpp
```diff
@@ -189,7 +189,22 @@ void UFlowAsset::HarvestNodeConnections()
 		}
 	}
 }
-#endif
+
+void UFlowAsset::AddCustomInput(const FName& InName)
```
LindyHopperGT, line 195: _Accessors for subclasses to use to modify the CustomInput array_

Source/Flow/Private/FlowSubsystem.cpp
```diff
@@ -265,7 +265,7 @@ TMap<UObject*, UFlowAsset*> UFlowSubsystem::GetRootInstances() const
 	{
 		Result.Emplace(RootInstance.Value.Get(), RootInstance.Key);
 	}
-	return Result;
```
LindyHopperGT: _Explicitly calling MoveTemp() to invoke move semantics.  (yes, compilers will often detect this optimization possibility, but I like being explicit, particularly when it involves a deep copy of an array)._

Source/Flow/Private/Nodes/FlowNode.cpp
```diff
@@ -148,6 +150,45 @@ UFlowAsset* UFlowNode::GetFlowAsset() const
 	return GetOuter() ? Cast<UFlowAsset>(GetOuter()) : nullptr;
 }
 
+AActor* UFlowNode::TryGetRootFlowActorOwner() const
```
LindyHopperGT, line 153: _This code moved from FlowNode_PlayLevelSequence into an UFlowNode function so that other FlowNodes can use it._

Source/Flow/Private/Nodes/Route/FlowNode_CustomInput.cpp
```diff
@@ -1,17 +1,14 @@
 // Copyright https://github.com/MothCocoon/FlowGraph/graphs/contributors
 
 #include "Nodes/Route/FlowNode_CustomInput.h"
+#include "FlowSettings.h"
+
+#define LOCTEXT_NAMESPACE "FlowNode"
 
 UFlowNode_CustomInput::UFlowNode_CustomInput(const FObjectInitializer& ObjectInitializer)
 	: Super(ObjectInitializer)
 {
```
LindyHopperGT, line 10: _Most of the CustomInput/Output code moved into CustomNodeBase, so this diff doesn't look so great (mostly removed a bunch of common stuff)_

Source/Flow/Private/Nodes/World/FlowNode_PlayLevelSequence.cpp
```diff
@@ -132,21 +132,7 @@ void UFlowNode_PlayLevelSequence::CreatePlayer()
 	{
 		ALevelSequenceActor* SequenceActor;
 
-		AActor* OwningActor = nullptr;
```
LindyHopperGT, line 135: _Moved to a function on UFlowNode_

Source/FlowEditor/Private/DetailCustomizations/FlowNode_CustomInputDetails.cpp
```diff
@@ -1,101 +1,35 @@
...
-#include "PropertyEditing.h"
-#include "Widgets/Input/SComboBox.h"
-#include "Widgets/Text/STextBlock.h"
 
 #define LOCTEXT_NAMESPACE "FlowNode_CustomInputDetails"
 
-void FFlowNode_CustomInputDetails::CustomizeDetails(IDetailLayoutBuilder& DetailLayout)
+FFlowNode_CustomInputDetails::FFlowNode_CustomInputDetails()
 {
-	DetailLayout.GetObjectsBeingCustomized(ObjectsBeingEdited);
```
LindyHopperGT, line 17: _For best diff, compare FlowNode_CustomInputDetails.h/cpp vs. FlowNode_CustomNodeBase.h/cpp._

_This is the most consequential code change in the changelist.  It restructures how some of the Details code works/stores its data to fix a bug with the ComboBox list contents._

Source/FlowEditor/Private/Graph/FlowGraphSchema.cpp
```diff
@@ -72,17 +74,48 @@ void UFlowGraphSchema::GetGraphContextActions(FGraphContextMenuBuilder& ContextM
...
+	FVector2D NodeOffset = FVector2D::ZeroVector;
+
 	// Start node
-	UFlowGraphNode* NewGraphNode = FFlowGraphSchemaAction_NewNode::CreateNode(&Graph, nullptr, UFlowNode_Start::StaticClass(), FVector2D::ZeroVector);
+	const bool bStartNodePlacedAsGhostNode = IsValid(AssetClassDefaults) ? AssetClassDefaults->bStartNodePlacedAsGhostNode : false;
+	UFlowGraphNode* StartGraphNode = CreateDefaultNode(Graph, AssetClassDefaults, UFlowNode_Start::StaticClass(), NodeOffset, bStartNodePlacedAsGhostNode);
+	check(IsValid(StartGraphNode));
+
+	// Add default nodes for all of the CustomInputs
+	if (IsValid(AssetClassDefaults))
```
LindyHopperGT, line 89: _Adding default nodes for any CustomInputs that exist upon creation of the UFlowAsset.

In the UFlowAsset class itself, this will do nothing, as it starts its life with no CustomInputs.  Subclasses, however, may add them in their constructor, and in that case, they'll show up as "default" nodes in the graph, below the Start node._

Source/FlowEditor/Private/DetailCustomizations/FlowNode_CustomNodeBaseDetails.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+	check(CachedEventNameSelected.IsValid());
+	EventNames.Add(CachedEventNameSelected);
+
+	if (ObjectsBeingEdited[0].IsValid() && ObjectsBeingEdited[0].Get()->GetOuter())
+	{
+		const UFlowAsset* FlowAsset = CastChecked<UFlowAsset>(ObjectsBeingEdited[0].Get()->GetOuter());
+		TArray<FName> SortedNames = BuildEventNames(*FlowAsset);
+
+		if (bExcludeReferencedEvents)
```
LindyHopperGT, line 71: _The contents of this "if" were the primary difference between the CustomInput and CustomOutput versions of this function.  The unified version uses this bool to choose whether it should do the CustomInput version (true) or the CustomOutput version (false)._

Source/FlowEditor/Private/DetailCustomizations/FlowNode_CustomNodeBaseDetails.cpp
```diff
@@ -0,0 +1,152 @@
...
+	ensure(ObjectsBeingEdited[0].IsValid());
+
+	UFlowNode_CustomNodeBase* Node = Cast<UFlowNode_CustomNodeBase>(ObjectsBeingEdited[0].Get());
+	if (IsValid(Node) && Item)
+	{
+		const bool bIsChanged = (*CachedEventNameSelected != *Item);
+
+		if (bIsChanged)
+		{
+			CachedEventNameSelected = Item;
```
LindyHopperGT, line 139: _We need to cache the selected item, and it also needs to be a member of the EventNames list (SComboBox code assumes that it is, and specifically the same TSharedPtr object, not just the value being the same)._

Source/FlowEditor/Private/DetailCustomizations/FlowNode_CustomNodeBaseDetails.cpp
```diff
@@ -0,0 +1,152 @@
...
+					[
+						SNew(STextBlock)
+							.Text(this, &FFlowNode_CustomNodeBaseDetails::GetSelectedEventText)
+					]
+			];
+}
+
+void FFlowNode_CustomNodeBaseDetails::OnComboBoxOpening()
+{
+	RebuildEventNames();
```
LindyHopperGT, line 56: _Calling this every time the ComboBox opens.  Previously was only called once when the details customization object was created, this was a problem if you switched the combo selection to a value, then tried to go back to the previous value, because it never updated the EventNames list._

Source/FlowEditor/Private/DetailCustomizations/FlowNode_CustomNodeBaseDetails.cpp
```diff
@@ -0,0 +1,152 @@
...
+	Category.AddCustomRow(CustomRowNameText)
+		.NameContent()
+			[
+				SNew(STextBlock)
+					.Text(EventNameText)
+			]
+		.ValueContent()
+			.HAlign(HAlign_Fill)
+			[
+				SAssignNew(EventTextListWidget, SComboBox<TSharedPtr<FName>>)
```
LindyHopperGT, line 42: _Storing a pointer to the widget so that we can refresh it later in **PinSelectionChanged**_

Source/FlowEditor/Private/DetailCustomizations/FlowNode_CustomNodeBaseDetails.cpp
```diff
@@ -0,0 +1,152 @@
...
+		{
+			const bool bIsCurrentSelection = (EventName == *CachedEventNameSelected);
+			if (!EventName.IsNone() && !bIsCurrentSelection)
+			{
+				EventNames.Add(MakeShared<FName>(EventName));
+			}
+		}
+	}
+
+	if (!IsInEventNames(NAME_None))
```
LindyHopperGT, line 97: _It's possible that the current selection is "None".  In that case only, the None is at the top of the combo box list.  This is consistent with the other cases where the 'current' selection is always forced to the top of the list._

Source/FlowEditor/Private/DetailCustomizations/FlowNode_CustomNodeBaseDetails.cpp
```diff
@@ -0,0 +1,152 @@
...
+{
+	// Subclasses must override this function (and call CustomizeDetailsInternal with the localized text)
+	checkNoEntry();
+}
+
+void FFlowNode_CustomNodeBaseDetails::CustomizeDetailsInternal(IDetailLayoutBuilder& DetailLayout, const FText& CustomRowNameText, const FText& EventNameText)
+{
+	DetailLayout.GetObjectsBeingCustomized(ObjectsBeingEdited);
+
+	if (ObjectsBeingEdited[0].IsValid())
```
LindyHopperGT, line 25: _Initially caching the current selection when the details customization object is first created.  This will be updated in **PinSelectionChanged**._

Source/Flow/Public/Nodes/Route/FlowNode_Start.h
```diff
@@ -13,6 +13,8 @@ class FLOW_API UFlowNode_Start : public UFlowNode
 {
 	GENERATED_UCLASS_BODY()
 
+	friend class UFlowAsset;
```
LindyHopperGT, line 16: _Added a friend here so that the UFlowGraph can call "ExecuteInput" rather than "TriggerOutputPin" directly.  It seemed more correct to execute the node rather than trigger its output directly.  The effect is the same for the code as it is now, but if the ExecuteInput has any additional logic added to it, the ExecuteInput would execute it properly, whereas the TriggerOutputPin version would have skipped the additional execute code._

_It could be made **public** instead of giving friend access.  Or you we revert the replacements for TriggerOutpinPin -> ExecuteInput and remove the need for the friend access._

Source/FlowEditor/Private/DetailCustomizations/FlowNode_CustomNodeBaseDetails.cpp
```diff
@@ -0,0 +1,152 @@
...
+			{
+				return *ExistingName == EventName;
+			});
+
+	return bIsInEventNames;
+}
+
+TSharedRef<SWidget> FFlowNode_CustomNodeBaseDetails::GenerateEventWidget(const TSharedPtr<FName> Item) const
+{
+	return
```
LindyHopperGT, line 116: _I reformatted this to be standard "Slate" coding style._

Source/FlowEditor/Private/DetailCustomizations/FlowNode_CustomOutputDetails.cpp
```diff
@@ -1,93 +1,35 @@
...
-						.OnSelectionChanged(this, &FFlowNode_CustomOutputDetails::PinSelectionChanged)
-						[
-							SNew(STextBlock)
-								.Text(this, &FFlowNode_CustomOutputDetails::GetSelectedEventText)
-						]
-				];
+	check(bExcludeReferencedEvents == false);
 }
 
-void FFlowNode_CustomOutputDetails::GetEventNames()
```
LindyHopperGT, line 41: _I renamed GetEventNames() to RebuildEventNames() because the "Get" naming implied it was a simple lookup (eg, accessor) but it actually did a lot of work and didn't return anything._

Source/FlowEditor/Private/Graph/FlowGraphSchema.cpp
```diff
@@ -72,17 +74,48 @@ void UFlowGraphSchema::GetGraphContextActions(FGraphContextMenuBuilder& ContextM
 
 void UFlowGraphSchema::CreateDefaultNodesForGraph(UEdGraph& Graph) const
 {
+	const UFlowAsset* AssetClassDefaults = GetAssetClassDefaults(&Graph);
+
+	static const FVector2D NodeOffsetIncrement = FVector2D(0, 128);
```
LindyHopperGT, line 79: _Offsetting by 128 puts each of the CustomInputs nicely on one of the Graph editor "thick" lines in its grid, which looks nice and tidy.  It also gives a decent (but not overly so) amount of room between each._

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -204,8 +206,6 @@ void UFlowAsset::RemoveCustomInput(const FName& InName)
 	CustomInputs.Remove(InName);
 }
 
-#endif // WITH_EDITOR
```
LindyHopperGT: These function bodies were incorrectly within the WITH_EDITOR ifdef block (causing non-WITH_EDITOR builds to fail if they used these functions).

Source/Flow/Private/FlowAsset.cpp
MothDoctor: 2 minor questions here
- Is it considered an editor-only code? I'd wrap it with `WITH_EDITOR`
- Would you see any problem with changing `checks` in Add/Remove method to simply ensure or even if? I tend to avoid potentially crashing checks if isn't necessary. Here nothing changes if a "wrong" event name would be provided.

MothDoctor:
Thanks for this big changelist! Merged! :)

- Renamed `UFlowNode_CustomNodeBase` to `UFlowNode_CustomEventBase` as this better conveys its functionality. Renamed Details Customizations class accordingly.
- A few minor formatting and readability tweaks to match conventions used in the plugin's codebase.
- Removed one line of stray code that you pointed out ;)

Source/Flow/Private/FlowAsset.cpp
LindyHopperGT: 1. AddCustomInput() could be Editor-only, if it's only ever expected to be done in editor.
2. I tend to use a lot of asserts (checks), but you could downgrade them to ensureRuntimeWarning or something.  ensure() is just a check() that is never stripped from the code, so it will still appear to crash.   Since you have your own messagelog in Flowgraph, you might write an error to there instead.

I'll be more careful to avoid check() in the future for this codebase.

Do you want me to include those changes in my next submission or will you make them?

MothDoctor: I'll do it.

However... what if I'd promoted the `CustomInputs` and `CustomOutputs` properties to protected? We wouldn't have to add these methods at all in the plugin. And it might be easier to maintain for you?

LindyHopperGT: Somebody on the discord made a good point about keeping them private and providing accessors. So I think it’s reasonable to keep the accessor pattern.

MothDoctor: Changes committed! Thanks again! :)

Issue #155, opened 2023-05-04
Play Level Sequence: "Pause at End" does nothing


soraphis:
I guess a temporary actor is created?
And when the flow node finishes, the actor seems to be destroyed which then resets the sequence.

My setup is kinda simple:
a box is moved from A to B in a sequence. After that it should remain at B.
It works when I create a LevelSequenceActor, but not when I use the "Play Level Sequence" node.

MothDoctor:
Should be fixed thanks to this PR :)
https://github.com/MothCocoon/FlowGraph/pull/156

PR #156, opened 2023-05-16
Play Level Sequence: fixed 'Pause at End' option.

twevs:
Closes issue #155.


MothDoctor:
Accepted, thanks for that! :)

PR #157, opened 2023-05-24
CustomOutput and quality of life changes

LindyHopperGT:
flowcomponent (subclasses) are now called when CustomOutputs are triggered for root flow graphs

Added a button for Edit Asset Defaults in the editor (clicking on the graph will also do this, but it's nice to have a button)

Added a flownode DeinitializeInstance call (called when its flowasset is deinitialized)


Flow.uplugin
```diff
@@ -1,29 +1,28 @@
 {
```
LindyHopperGT, line 1: I didn't intentionally change this file, but I couldn't figure out how to get GIT to keep it out of the PR.

(We're running UE 5.1 still, so we have to have a different version of this file locally).

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -189,35 +189,73 @@ void UFlowAsset::HarvestNodeConnections()
 	}
 }
 
-void UFlowAsset::AddCustomInput(const FName& InName)
+void UFlowAsset::AddCustomInput(const FName& EventName)
```
LindyHopperGT, line 192: Renamed the parameter to align it with the CustomOutput version

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -189,35 +189,73 @@ void UFlowAsset::HarvestNodeConnections()
...
+void UFlowAsset::AddCustomOutput(const FName& EventName)
+{
+	check(!CustomOutputs.Contains(EventName));
+
+	CustomOutputs.Add(EventName);
+}
+
+void UFlowAsset::RemoveCustomOutput(const FName& EventName)
+{
+	check(CustomOutputs.Contains(EventName));
```
LindyHopperGT, line 217: Any code calling these functions should not be double-adding or double removing custom events, so I'm asserting it as a code error.

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -189,35 +189,73 @@ void UFlowAsset::HarvestNodeConnections()
...
+
+	CustomOutputs.Remove(EventName);
+}
 #endif // WITH_EDITOR
 
+UFlowNode_CustomInput* UFlowAsset::TryFindCustomInputNodeByEventName(const FName& EventName) const
```
LindyHopperGT, line 223: New (public) function to find a custom input by name.  

We use it in our code to look to see if the flowgraph has specific inputs, and check if they have connected nodes, before attempting to call them.

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -189,35 +189,73 @@ void UFlowAsset::HarvestNodeConnections()
...
+}
+
 UFlowNode* UFlowAsset::GetDefaultEntryNode() const
 {
+	UFlowNode* FirstStartNode = nullptr;
+
 	for (const TPair<FGuid, UFlowNode*>& Node : Nodes)
 	{
 		UFlowNode_Start* StartNode = Cast<UFlowNode_Start>(Node.Value);
-		if (StartNode && StartNode->GetConnectedNodes().Num() > 0)
```
LindyHopperGT, line 214: split this if so we can cache the FirstStartNode (see later comment)

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -189,35 +189,73 @@ void UFlowAsset::HarvestNodeConnections()
...
+			{
+				FirstStartNode = StartNode;
+			}
 		}
 	}
 
-	return nullptr;
+	// If none of the found start nodes have connections, 
+	//  fallback to the first start node we found
+	return FirstStartNode;
```
LindyHopperGT, line 258: If none of the start nodes had any connected nodes (which is often the case for our graphs) then falling back to the first start node discovered.

We only use "Start" node for one-time initialization.  Most of the connected nodes are connected to custom inputs, which aren't Start nodes. 

After integrating with latest, but without this fix, our flow graphs never started (because nothing was connected to the start node).

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -326,6 +364,17 @@ void UFlowAsset::InitializeInstance(const TWeakObjectPtr<UObject> InOwner, UFlow
 
 void UFlowAsset::DeinitializeInstance()
 {
+	for (auto& KV : Nodes)
```
LindyHopperGT, line 367: Cleanup() is called when the node is "Finished", but we want to do some cleaning up only when the entire flow asset graph is shutdown.  

Adding this call for each of the flownodes to do this work when the FlowAsset's DeinitializeInstance() is called

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -421,34 +470,76 @@ void UFlowAsset::FinishFlow(const EFlowFinishPolicy InFinishPolicy, const bool b
 	}
 }
 
+bool UFlowAsset::HasStartedFlow() const
```
LindyHopperGT, line 473: I couldn't find a way to ask "Hey, has this flow graph started yet?", so added this.  

There might be a better way to answer the question, but this one works for my needs (so far).

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -421,34 +470,76 @@ void UFlowAsset::FinishFlow(const EFlowFinishPolicy InFinishPolicy, const bool b
...
+{
+	return RecordedNodes.Num() > 0;
+}
+
+AActor* UFlowAsset::TryFindActorOwner() const
```
LindyHopperGT, line 478: This will return the owning actor if the flow graph is owned by a component.

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -421,34 +470,76 @@ void UFlowAsset::FinishFlow(const EFlowFinishPolicy InFinishPolicy, const bool b
...
+
+	return nullptr;
+}
+
 TWeakObjectPtr<UFlowAsset> UFlowAsset::GetFlowInstance(UFlowNode_SubGraph* SubGraphNode) const
 {
 	return ActiveSubGraphs.FindRef(SubGraphNode);
 }
 
-void UFlowAsset::TriggerCustomEvent(UFlowNode_SubGraph* Node, const FName& EventName) const
```
LindyHopperGT, line 429: Because I changed around the custom input/output event calls, I made them more clear which versions were subgraph-specific and which are called for root graphs as well.

Previously, CustomOutputs wouldn't get called on a root graph, now they are, so some disambiguation on the function naming was necessary to keep the code clear.

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -421,34 +470,76 @@ void UFlowAsset::FinishFlow(const EFlowFinishPolicy InFinishPolicy, const bool b
...
 }
 
-void UFlowAsset::TriggerCustomOutput(const FName& EventName) const
+void UFlowAsset::TriggerCustomOutput(const FName& EventName)
 {
-	NodeOwningThisAssetInstance->TriggerOutput(EventName);
+	check(HasStartedFlow());
+
+	const bool bIsSubgraph = NodeOwningThisAssetInstance.IsValid();
+	if (bIsSubgraph)
```
LindyHopperGT, line 525: Here is the inflection point where we change where the output trigger information goes to.

For Subgraphs, it does the previous behavior.

For root graphs it dispatches the event to the owning component.

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -421,34 +470,76 @@ void UFlowAsset::FinishFlow(const EFlowFinishPolicy InFinishPolicy, const bool b
...
+		if (UFlowComponent* FlowComponent = Cast<UFlowComponent>(GetOwner()))
+		{
+			FlowComponent->OnTriggerRootFlowOutputEventDispatcher(*this, EventName);
+		}
+	}
 }
 
 void UFlowAsset::TriggerInput(const FGuid& NodeGuid, const FName& PinName)
 {
+	check(HasStartedFlow());
```
LindyHopperGT, line 541: I wanted to catch any cases where TriggerInput() was called prior to StartFlow() being called.

Source/Flow/Private/Nodes/Route/FlowNode_CustomOutput.cpp
```diff
@@ -14,9 +15,44 @@ UFlowNode_CustomOutput::UFlowNode_CustomOutput(const FObjectInitializer& ObjectI
 
 void UFlowNode_CustomOutput::ExecuteInput(const FName& PinName)
 {
-	if (!EventName.IsNone() && GetFlowAsset()->GetCustomOutputs().Contains(EventName) && GetFlowAsset()->GetNodeOwningThisAssetInstance())
+	UFlowAsset* FlowAsset = GetFlowAsset();
+	check(IsValid(FlowAsset));
+
+	if (EventName.IsNone())
```
LindyHopperGT, line 21: Added some logged warnings for bad configuration or mismatched parameters (would fail silently before).

Source/Flow/Public/FlowComponent.h
```diff
@@ -203,9 +203,23 @@ class FLOW_API UFlowComponent : public UActorComponent
 	UFUNCTION(BlueprintPure, Category = "RootFlow", meta = (DeprecatedFunction, DeprecationMessage="Use GetRootInstances() instead."))
 	UFlowAsset* GetRootFlowInstance() const;
 
+//////////////////////////////////////////////////////////////////////////
+// UFlowComponent overrideable events
+
+public:
+	// Called when a Root flow asset triggers a CustomOutput
+	UFUNCTION(BlueprintImplementableEvent, DisplayName = "OnTriggerRootFlowOutputEvent")
```
LindyHopperGT, line 211: Blueprint overrideable event.  The dispatcher will still call the blueprint and code versions (so that the blueprint author does not need to 'remember' to call the parent version in addition to their own).

Source/Flow/Public/FlowComponent.h
```diff
@@ -203,9 +203,23 @@ class FLOW_API UFlowComponent : public UActorComponent
...
+	void BP_OnTriggerRootFlowOutputEvent(UFlowAsset* RootFlowInstance, const FName& EventName);
+	virtual void OnTriggerRootFlowOutputEvent(UFlowAsset& RootFlowInstance, const FName& EventName) { }
```
LindyHopperGT, line 213: This is the code overrideable.  Any flow component that wants to handle output events can override this in code (and/or the blueprint version above)

Source/Flow/Public/FlowComponent.h
```diff
@@ -203,9 +203,23 @@ class FLOW_API UFlowComponent : public UActorComponent
...
+
+	//~Begin UFlowAsset-only access
+	FORCEINLINE void OnTriggerRootFlowOutputEventDispatcher(UFlowAsset& RootFlowInstance, const FName& EventName);
```
LindyHopperGT, line 216: Dispatcher calls both the blueprint and code overrides for the output trigger event

Source/Flow/Public/Nodes/FlowNode.h
```diff
@@ -332,6 +332,10 @@ class FLOW_API UFlowNode : public UObject, public IVisualLoggerDebugSnapshotInte
 	UFUNCTION(BlueprintImplementableEvent, Category = "FlowNode", meta = (DisplayName = "Cleanup"))
 	void K2_Cleanup();
 
+	// Method called from UFlowAsset::DeinitializeInstance() 
+	//  (ie, when deinitializing the flow asset itself)
+	virtual void DeinitializeInstance() { }
```
LindyHopperGT, line 337: Base-class DeinitializeInstance function for flownode does nothing (but subclasses might want to do stuff)

Source/FlowEditor/Private/Asset/FlowAssetEditor.cpp
```diff
@@ -358,6 +358,10 @@ void FFlowAssetEditor::BindToolbarCommands()
 								FExecuteAction::CreateSP(this, &FFlowAssetEditor::ValidateAsset_Internal),
 								FCanExecuteAction());
 
+	ToolkitCommands->MapAction(ToolbarCommands.EditAssetDefaults,
```
LindyHopperGT, line 361: Adding the Edit Asset Defaults button

Source/FlowEditor/Private/Asset/SFlowDiff.cpp
```diff
@@ -6,12 +6,17 @@
 #include "FlowAsset.h"
 
 #include "EdGraphUtilities.h"
+#include "Editor.h"
```
LindyHopperGT, line 9: Compiling this in UE5.1 had a lot of missing headers for some reason.  

I filled them in.

Flow.uplugin
MothDoctor: Umh... so just reverted change after merging PR...

Source/Flow/Private/FlowAsset.cpp
MothDoctor: I understand motivation, but I'd like to keep consistent with the rest of the plugin's code so... I converted checks to regular If statements.  Might review this approach later ;)

Source/Flow/Private/FlowAsset.cpp
MothDoctor: Very useful fix, thanks :)

Source/Flow/Private/FlowAsset.cpp
MothDoctor: I'm sorry, but I'm removing all these checks which might crash in the editor (leading to lost work of a given developer) or even in released games in some edge cases.
I'd recommend using here at least logging which would inform a non-programmer what happened. I spent too much time in life on resolving crash reports caused by issued that might easily fixed by designer alone :)

I understand the need of adding checks everywhere to catch invalid data/events during development. Might accept `check` in runtime code, if it was optionally compiled out :)

Source/Flow/Public/Nodes/FlowNode.h
MothDoctor: Nice one :)
Added here calling a blueprint event `K2_DeinitializeInstance` by default. I suspect that someone would request it at some point.

Source/Flow/Private/FlowAsset.cpp
MothDoctor: Sure. I admit I'm puzzled about what would be the best name for the function calling Custom Event from SubGraph... For now, I'm settled on `TriggerCustomInput_FromSubGraph`. Let me know if my version sounds worse to your ears ;)

Source/Flow/Private/FlowAsset.cpp
MothDoctor: Neat!
Only tweaking comments here :)

Source/Flow/Private/FlowAsset.cpp
MothDoctor: Interesting "problem", indeed ;)

Source/FlowEditor/Private/Asset/FlowAssetEditor.cpp
MothDoctor: Now when I'm seeing it... I wonder how we could live without it? xD

MothDoctor: Update: moved the Asset Defaults button after Diff and Search buttons, this is more consistent with a blueprint toolbar :)

MothDoctor:
Awesome, there are a lot of good changes introduced with a small amount of code! It's all clean, so I only tweaked a few places to improve code readability and consistency (satisfying my styling OCD).

Any non-cosmetic changes described in the conversations. However, there's nothing changing your functionality :)
Big thanks!

PS Usually, I would ask for re-creating PR if combines many things at once. But I see you fought a battle with Git here ;)

Source/Flow/Private/FlowAsset.cpp
LindyHopperGT: Sounds fine.  I just needed to disambiguate between the subgraph version and the general (root or subgraph) version.

Source/Flow/Private/FlowAsset.cpp
LindyHopperGT: in UE stock, "check" is not included in Test or Shipping configurations.  "ensure" is included in all versions.

[https://docs.unrealengine.com/4.26/en-US/ProgrammingAndScripting/ProgrammingWithCPP/Assertions/](url)

For my team, we use check() for programmer errors, to require that programmers are calling the code as expected.  For any errors that can be caused by blueprint or non-programmers, we use the unreal error reporting tools (often MessageLog or UE_LOG).  This sort of check is appropriate according to my code standard, because it polices **code** calling the functions incorrectly, and would be found and fixed by a programmer before ever being put in front of an artist or designer's eyes.

All that said, I don't mind you stripping them out here, it's ultimately your code.

Source/FlowEditor/Private/Asset/FlowAssetEditor.cpp
LindyHopperGT: Makes sense.

LindyHopperGT:
Yeah, I'm still bumbling my way through GIT.  It's a pretty different mindset than the source control packages I've used before (CVS, VSS, SVN, P4, etc.)

Future changes will be more localized commits.  this was an omnibus of the tweaks for the last few weeks.  Sorry it got all mingled together.

PR #160, opened 2023-05-31
Updated code to build in 5.1 and 5.2

LindyHopperGT:
Added some #if's around code that changed from 5.1 to 5.2 to allow it to compile in both versions (presumably also 5.0, but that's untested).

Additionally to build with 5.1, the **flow.uplugin** needs to be locally modified to point to 5.1 instead of 5.2

### DIFF
--- a/Source/Flow/Private/FlowAsset.cpp
+++ b/Source/Flow/Private/FlowAsset.cpp
@@ -31,6 +31,8 @@ UFlowAsset::UFlowAsset(const FObjectInitializer& ObjectInitializer)
 	{
 		AssetGuid = FGuid::NewGuid();
 	}
+
+	ExpectedOwnerClass = UFlowSettings::Get()->GetDefaultExpectedOwnerClass();
 }
 
 #if WITH_EDITOR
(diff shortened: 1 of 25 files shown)


LindyHopperGT:
I have the Call Owner Function request ready, but waiting for this one to clear first.

MothDoctor:
@LindyHopperGT Mhm, actually Call Owner Function got added to this PR ;)
I see no problem, could we just upgrade this PR to be about Call Owner Function? UE 5.2 version checks can stay here, it's a very small change :)

I just looked quickly at the code. Wonder if we could merge the `UnrealExtensions` and `DetailCustomizations` folders into single `Customizations`? Later I would simply rename the old `DetailCustomizations` folder into `Customizations`. I think it would be simply confusing to occasional users to see multiple customization folders. What do you think?

LindyHopperGT:
Sure. I’m Not sure how that happened.  Git is a mystery to me I’m finishing up my vacation in Japan. Will be back on my work pc in a few days. Cheers,Greg TaylorOn Jul 9, 2023, at 1:23 AM, Krzysiek Justyński ***@***.***> wrote:﻿
Merged #160 into 5.2.

PR #161, opened 2023-06-06
Add class hyperlink

Drakynfly:
This PR adds a class hyperlink to the menu in a similar fashion to how BP's and other assets do. This is mostly to help when subclassing FlowAsset, and being able to tell at a glance which class you are in.


Drakynfly:
image

This is what it looks like btw.

MothDoctor:
@Drakynfly Accepted it, thanks! :)

PR #162, opened 2023-06-12
Play Level Sequence: added Pause and Resume input pins.

twevs:
As mentioned by Mael on Discord. :) (Submitting on his behalf as I handle our Git stuff.)


MothDoctor:
@twevs Accepted, thanks!
I only moved `Stop` input to the bottom :)

PR #164, opened 2023-06-21
Fix JumpToInnerObject not working when SearchTree provides the GraphNode itself

growlitheharpo:
After making the corresponding engine changes, I found that - depending on the exact search term that was entered - the SearchTree would sometimes provide the GraphNode instead of the FlowNode for a particular node. As a result, the focus failed.

This just adds handling for when that happens and I have not had any further issues with focus.


MothDoctor:
Awesome, welcome to contributors! :)

PR #165, opened 2023-06-22
Add validation error on FlowAsset for disallowed node classes

Deams51:
Validation was not checking for disallowed node classes in a flow asset, which can happen if the node classes allowed were changed after nodes were added to the graph. 

To not duplicate the node class allowed logic, I've encapsulated it in a function on the flow asset and it is being called by the validation and schema.


MothDoctor:
Accepted, thanks!
It's nice moving existing code from the Graph Schema to the asset itself :)

PR #166, opened 2023-06-22
Add ability for FlowAsset child classes to define if they are allowed in subgraph nodes

Deams51:
~~To prevent custom FlowAsset classes to be instantiated by a subgraph node, this PR adds:~~
~~- A virtual method that returns if the class can be used in subgraph nodes~~
~~- Manually setting the **DisallowedClasses** meta specifier on the subgraph node Asset UProperty~~
~~- A validation rule on subgraph nodes~~


Deams51:
I do see an issue with this implementation when creating a child class of `UFlowNode_SubGraph`.
Will push a new version with a better architecture.

Deams51:
Ready for review. 

This PR now adds the ability to choose which `UFlowAsset` classes can be used by `UFlowNode_SubGraph` classes so that custom flow assets instantiation can be restricted on custom assets to specific start nodes. 
The restriction is purely editor-side and prevents the selection of the disallowed classes in the class picker.

PR #167, opened 2023-06-22
Consolidate settings

ryanjon2040:
Before Flow settings were scattered around and was hard to find so I've organized them into one place. Now you can find them in: 

**Project Settings**
image

**Editor Settings**
image

Previously they were here:
```
Editor Preferences -> General -> Flow Graph
Project Settinngs -> Editor -> Flow Graph
Project Settings -> Game -> Flow
```

### DIFF
--- a/Source/Flow/Public/FlowSettings.h
+++ b/Source/Flow/Public/FlowSettings.h
@@ -42,4 +42,12 @@ class FLOW_API UFlowSettings : public UDeveloperSettings
 	// by incorporating data that would otherwise go in the Description
 	UPROPERTY(EditAnywhere, config, Category = "Nodes")
 	bool bUseAdaptiveNodeTitles;
+
+public:
+
+#if WITH_EDITORONLY_DATA
+	virtual FName GetCategoryName() const override { return FName("Flow Graph"); }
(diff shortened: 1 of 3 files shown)


ryanjon2040:
@MothDoctor I think `FlowGraphEditorSettings.h` is redundant and all those settings makes more sense in `FlowGraphSettings.h`. I thought of deleting it and moving all those settings to the other one but not sure if it is something you would merge.

MothDoctor:
@ryanjon2040 `FlowGraphEditorSettings.h` isn't redundant, as it contains personal per-user settings :)
I understand the confusion here, I was thinking a few times to rename this class to `FlowGraphUserSettings.h`. But I need to check if core redirect would work, so nobody would loose their personal settings.

ryanjon2040:
> @ryanjon2040 `FlowGraphEditorSettings.h` isn't redundant, as it contains personal per-user settings :) I understand the confusion here, I was thinking a few times to rename this class to `FlowGraphUserSettings.h`. But I need to check if core redirect would work, so nobody would loose their personal settings.

Flow Graph user settings makes more sense IMHO.

PR #174, opened 2023-09-26
Fixed crash when doing a diff of a Flow Graph vs Perforce.

InfiniteAutomaton:
The crash occurred inside of the SetContent method called from SFlowDiff::SetCurrentMode. When the Widget is not set to SNullWidget::NullWidget it assumes that the widget exists and attempts to dereference it, but in the former code the Widget was unset (nullptr) causing the crash.


MothDoctor:
Thanks @InfiniteAutomaton! Accepted :)

PR #177, opened 2023-10-28
Add EFlowNodeDoubleClickTarget::PrimaryAssetOrNodeDefinition to support opening the Asset and falling back to opening the Node Class if the node doesn't have an asset.

DoubleDeez:


MothDoctor:

PR #178, opened 2023-11-12
Non-unity mode build fixes

InfiniteAutomaton:
I needed to switch my project to non-unity mode. These are the changes required to keep FlowGraph compiling in that configuration.


MothDoctor:

Issue #172, opened 2023-08-16
Alignment feature does not work in the FlowGraph


samaursa:
Unreal's node alignment feature and the alignment shortcuts (`Shift+A`, `Shift+D`, `Shift+S` etc.) do not work in the NodeGraph. It would be great if this feature was added to the graph:

![image](https://github.com/MothCocoon/FlowGraph/assets/817849/c812a805-062c-4b81-8170-3c0e4414fa20)

MothDoctor:
Feature implemented by @VintageDeveloper :)
https://github.com/MothCocoon/FlowGraph/pull/179

PR #179, opened 2023-11-22
Fix Issue #172 - Add Alignment Features to Flow Graph Nodes

VintageDeveloper:
**Pull Request** 
This Pull Request makes changes based on feedback from Issue #172.

**Key Change**
Adds Context Menu Actions, "Alignment", for Flow Graph Nodes. Can be accessed similar to Alignments in Blueprints:
- Right Click Flow Graph Node and select from the Alignments
- Using Alignment Shortcuts


MothDoctor:
Merged that, thanks a lot! :)

PR #180, opened 2023-11-29
Additions, refinement and bugfixes

LindyHopperGT:
- Fixed a bug in CallOwnerFunction where it would not bind correctly.
- Fixed an bug where flow nodes were in the palette, even if they were in inaccessible plugins (now FlowNode palette respects plugin access rules)
- Added expanded CustomOutput access
- Clarified editor-accessible vs. runtime accessible information for custom in/outputs
- Split out some functionality for subsystem registration to be called separately (to support our object pooling model)


Source/Flow/Private/FlowAsset.cpp
```diff
@@ -71,16 +76,45 @@ void UFlowAsset::PostDuplicate(bool bDuplicateForPIE)
 	}
 }
 
+void UFlowAsset::PostLoad()
+{
+	Super::PostLoad();
+
+	// If we removed or moved a flow node blueprint (and there is no redirector) we might loose the reference to it resulting
```
LindyHopperGT, line 83: This code fixes-up flow assets that were corrupted by the fixed bug in CallOwnerFunction.

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -71,16 +76,45 @@ void UFlowAsset::PostDuplicate(bool bDuplicateForPIE)
...
 {
 	// validate nodes
 	for (const TPair<FGuid, UFlowNode*>& Node : Nodes)
 	{
-		if (Node.Value)
+		if (IsValid(Node.Value))
 		{
-			if (!IsNodeClassAllowed(Node.Value->GetClass()))
+			FText FailureReason;
+			if (!IsNodeClassAllowed(Node.Value->GetClass(), &FailureReason))
```
LindyHopperGT, line 111: Validate will now report inaccessible flow nodes being used in a flow graph (that is, they're in an inaccessible plugin).

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -90,74 +124,124 @@ EDataValidationResult UFlowAsset::ValidateAsset(FFlowMessageLog& MessageLog)
...
-bool UFlowAsset::IsNodeClassAllowed(const UClass* FlowNodeClass) const
+bool UFlowAsset::IsNodeClassAllowed(const UClass* FlowNodeClass, FText* OutOptionalFailureReason) const
 {
-	if (FlowNodeClass == nullptr)
+	if (!IsValid(FlowNodeClass))
+	{
+		return false;
+	}
+
+	if (!CanFlowNodeClassBeUsedByFlowAsset(*FlowNodeClass))
```
LindyHopperGT, line 144: We broke-out these checks into their own subfunctions (below) and added a new one to check plugin access rules for the node, relative to this flow graph.

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -90,74 +124,124 @@ EDataValidationResult UFlowAsset::ValidateAsset(FFlowMessageLog& MessageLog)
...
+	if (!CanFlowAssetReferenceFlowNode(*FlowNodeClass, OutOptionalFailureReason))
 	{
 		return false;
 	}
 
-	UFlowNode* NodeDefaults = FlowNodeClass->GetDefaultObject<UFlowNode>();
+	return true;
+}
+
+bool UFlowAsset::CanFlowNodeClassBeUsedByFlowAsset(const UClass& FlowNodeClass) const
```
LindyHopperGT, line 163: We split-out some of the pre-existing checks into single-topic functions to improve code clarity.

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -341,6 +425,56 @@ UFlowNode_CustomInput* UFlowAsset::TryFindCustomInputNodeByEventName(const FName
 	return nullptr;
 }
 
+UFlowNode_CustomOutput* UFlowAsset::TryFindCustomOutputNodeByEventName(const FName& EventName) const
```
LindyHopperGT, line 428: Access function for custom output by event name.

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -341,6 +425,56 @@ UFlowNode_CustomInput* UFlowAsset::TryFindCustomInputNodeByEventName(const FName
...
+			{
+				return CustomOutput;
+			}
+		}
+	}
+
+	return nullptr;
+}
+
+TArray<FName> UFlowAsset::GatherCustomInputNodeEventNames() const
```
LindyHopperGT, line 444: The "name" versions of the custom input/output are editor-only, so if the names of inputs/outputs are used at runtime, we need to gather them from the nodes list.  These functions handle that work.

Source/Flow/Private/FlowComponent.cpp
```diff
@@ -44,6 +44,11 @@ void UFlowComponent::BeginPlay()
 {
 	Super::BeginPlay();
 
+	RegisterWithFlowSubsystem();
```
LindyHopperGT, line 47: Our object pooling model required different timing for registration with the subsystem than strictly in Begin/End Play, so we split these parts out into dedicated functions

Source/Flow/Private/Nodes/World/FlowNode_CallOwnerFunction.cpp
```diff
@@ -368,27 +387,20 @@ UClass* UFlowNode_CallOwnerFunction::GetParamsClassForFunctionName(const UClass&
...
 		return false;
 	}
 
 	const UClass* ExistingParamsClass = GetExistingParamsClass();
+	UClass* RequiredParamsClass = GetRequiredParamsClass();
 
 	const bool bNeedsAllocateParams = 
 		!IsValid(ExistingParamsClass) ||
-		!ExistingParamsClass->IsChildOf(RequiredParamsClass);
+		ExistingParamsClass != RequiredParamsClass;
```
LindyHopperGT, line 403: Requiring explicit class match now.  This is required by the new approach for allocating the correct params object for a CallOwner function.

Source/Flow/Private/Nodes/World/FlowNode_CallOwnerFunction.cpp
```diff
@@ -97,12 +97,31 @@ void UFlowNode_CallOwnerFunction::PostLoad()
...
+	//  (it could be removed in a future release, once all assets have been updated)
+	if (ParamsProperty->PropertyClass != UFlowOwnerFunctionParams::StaticClass())
+	{
+		ParamsProperty->SetPropertyClass(UFlowOwnerFunctionParams::StaticClass());
+	}
+}
+
+bool UFlowNode_CallOwnerFunction::CanEditChange(const FProperty* InProperty) const
+{
+	if (!Super::CanEditChange(InProperty))
```
LindyHopperGT, line 112: We're not allowing the user to choose the params object directly any longer (the way this was done previously was dangerous).  So now the params object is only created when selecting the function, and this CanEditChange prevents the user from choosing an incompatible params object.

Source/Flow/Public/FlowAsset.h
```diff
@@ -210,6 +206,25 @@ class FLOW_API UFlowAsset : public UObject
 		}
 	}
 
+public:
```
LindyHopperGT, line 209: Regrouped and expanded the custom input/output access functions

Source/FlowEditor/Private/Asset/FlowAssetToolbar.cpp
```diff
@@ -202,6 +202,7 @@ void FFlowAssetToolbar::BuildAssetToolbar(UToolMenu* ToolbarMenu) const
 		// add buttons
 		Section.AddEntry(FToolMenuEntry::InitToolBarButton(FFlowToolbarCommands::Get().RefreshAsset));
 		Section.AddEntry(FToolMenuEntry::InitToolBarButton(FFlowToolbarCommands::Get().ValidateAsset));
+		Section.AddEntry(FToolMenuEntry::InitToolBarButton(FFlowToolbarCommands::Get().EditAssetDefaults));
```
LindyHopperGT, line 205: Moved this section to align with the others

Source/FlowEditor/Private/DetailCustomizations/FlowOwnerFunctionRefCustomization.cpp
```diff
@@ -69,8 +69,19 @@ TArray<FName> FFlowOwnerFunctionRefCustomization::GetFlowOwnerFunctionRefs(
 	TArray<FName> ValidFunctionNames;
 
 	// Gather a list of potential functions
-	TArray<FName> PotentialFunctionNames;
-	ExpectedOwnerClass.GenerateFunctionList(PotentialFunctionNames);
+	TSet<FName> PotentialFunctionNames;
+
+	const UClass* CurClass = &ExpectedOwnerClass;
```
LindyHopperGT, line 74: Needed to gather functions from superclasses, and not just the leaf class.

Source/FlowEditor/Private/Graph/FlowGraph.cpp
```diff
@@ -31,7 +31,13 @@ UFlowGraph::UFlowGraph(const FObjectInitializer& ObjectInitializer)
 
 UEdGraph* UFlowGraph::CreateGraph(UFlowAsset* InFlowAsset)
 {
-	UEdGraph* NewGraph = CastChecked<UFlowGraph>(FBlueprintEditorUtils::CreateNewGraph(InFlowAsset, NAME_None, StaticClass(), UFlowGraphSchema::StaticClass()));
+	return CreateGraph(InFlowAsset, UFlowGraphSchema::StaticClass());
+}
+
+UEdGraph* UFlowGraph::CreateGraph(UFlowAsset* InFlowAsset, TSubclassOf<UFlowGraphSchema> FlowSchema)
```
LindyHopperGT, line 37: We use our own Schema (subclass) so we needed a CreateGraph function that accepted a user-specified schema.

We may want to remove the other version of this function (w/o the explicit schema param), but I didn't make that change in this changelist.

Source/FlowEditor/Private/Graph/FlowGraphSchema.cpp
```diff
@@ -54,9 +54,9 @@ void UFlowGraphSchema::SubscribeToAssetChanges()
 	}
 }
 
-void UFlowGraphSchema::GetPaletteActions(FGraphActionMenuBuilder& ActionMenuBuilder, const UClass* AssetClass, const FString& CategoryName)
+void UFlowGraphSchema::GetPaletteActions(FGraphActionMenuBuilder& ActionMenuBuilder, const UFlowAsset* EditedFlowAsset, const FString& CategoryName)
```
LindyHopperGT, line 57: This is reworked to not use the default flow asset, but instead use the actual flow asset (that is being edited) so we can ask the right class about which nodes are legal for the palette.

Source/FlowEditor/Public/DetailCustomizations/FlowOwnerFunctionRefCustomization.h
```diff
@@ -48,6 +48,4 @@ class FFlowOwnerFunctionRefCustomization : public IFlowCuratedNamePropertyCustom
 	static TArray<FName> GetFlowOwnerFunctionRefs(const UFlowNode_CallOwnerFunction& FlowNodeOwner, const UClass& ExpectedOwnerClass);
 
 	static bool IsFunctionUsable(const UFunction& Function, const UFlowNode_CallOwnerFunction& FlowNodeOwner);
-	static bool DoesFunctionHaveExpectedParamType(const UFunction& Function, const UFlowNode_CallOwnerFunction& FlowNodeOwner);
```
LindyHopperGT, line 51: the fixed model for CallOwnerFunction configuration does not need this function any longer.

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -90,74 +124,124 @@ EDataValidationResult UFlowAsset::ValidateAsset(FFlowMessageLog& MessageLog)
...
+		if (!bAllowedInAsset)
+		{
+			return false;
+		}
+	}
+
+	return true;
+}
+
+bool UFlowAsset::CanFlowAssetReferenceFlowNode(const UClass& FlowNodeClass, FText* OutOptionalFailureReason) const
```
LindyHopperGT, line 227: This is the new check if the node is accessible from the plugin context of the flow asset.

MothDoctor:
Merged that, thanks a lot again! :)

PR #184, opened 2024-01-12
fixes missing indentation in the palette panel for subcategories

### LINKED ISSUE #181: Palette does not indent headers for sub-categories
UnrealEditor_MJ7NoBlfif
_the header for the "!Actions|Player" sub categorie is not indented_

soraphis:
using a custom "RowExpander" Delegate analogous to the way the blueprint graph does it.
  compare it to here: https://github.com/EpicGames/UnrealEngine/blob/072300df18a94f18077ca20a14224b5d99fee872/Engine/Source/Editor/Kismet/Private/SBlueprintSubPalette.cpp#L293C54-L293C74

fixes #181


MothDoctor:
Awesome! Thanks for the fix :)

PR #185, opened 2024-01-22
Update FlowGraph Reroute Node Selection & Connection Highlighting to be similar to Blueprints. Resolves #171

### LINKED ISSUE #171: Reroute node does not format correctly when going in reverse
I have a dialogue system and sometimes a dialogue option can lead to other dialogue, which loops back to the original dialogue option. But the reroute node seems to not be functioning correctly or not formatting correctly, at least when compared to the blueprint reroute node.

Current behavior:
image

Blueprint behavior:
image

VintageDeveloper:
**Change 1:** When mouse cursor moves over a connection, the connection now highlights. This is similar to how connection highlighting is done in Blueprints.

image

**Change 2:** From Feedback #171, when reroute nodes are selected, in a reverse loop, the loop does not reverse. This is due to reroute nodes, when selected, also highlighting the "in" connection, leading to the loop not reversing. This pull request changes reroute node selection to not highlight the "in" connection - similar to how reroute node selection is done in Blueprints.

image

### DIFF
--- a/Source/FlowEditor/Private/Graph/FlowGraphConnectionDrawingPolicy.cpp
+++ b/Source/FlowEditor/Private/Graph/FlowGraphConnectionDrawingPolicy.cpp
@@ -196,6 +196,14 @@ void FFlowGraphConnectionDrawingPolicy::DetermineWiringStyle(UEdGraphPin* Output
 			}
 		}
 	}
+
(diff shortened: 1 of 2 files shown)


MothDoctor:
Oh, it's great you fixed that annoying visual issue! Accepted :)

PR #186, opened 2024-01-29
Fix UFlowNode_SubGraph in cooked builds where Asset member could dere…

jnucc:
…ference to nullptr, when trying to get the PathName after converting to UObject*. Just try to get the Path Name from the SoftObjectPtr itself instead.

Reason: UObjectBaseUtility::GetPathName() has a check with a if (this != NULL) and under clang, in optimized builds, this is assumed to be always true. Depending on how inlining goes, the compiler is free to eliminate if it assumes it will always be true; chcked generated UE 5.2 and UE 5.3 code in disassembly, and the checks for if(this != StopOuter && this != NULL) was outright eliminated during aggressive inlining and was assumed to be true, where in UE 5.2 there were still function calls and it cannot assume stuff can be nullptr or not.

On cooked builds, Asset could be not yet loaded.


MothDoctor:
Amazing! Thanks for the fix and detailed explanation :)

PR #188, opened 2024-03-03
Some improvements for working with CallOwnerFuntion

MaksymKapelianovych:
1. Display node description above node (as for triggers, timers etc).
2. Support for context pins (as for PlayLevelSequence). Now if Params type is changed in referenced function or pins inside Params are renamed/added/removed, CallOwnerFuntion node will not be refreshed. Pins will appear only if user changes referenced function to some other function and then returns back, which is not very convenient. After my changes, Refresh action will update node pins as well. Also, deleted some now excessive function to unify logic for pins refreshing.

### DIFF
--- a/Source/Flow/Private/FlowModule.cpp
+++ b/Source/Flow/Private/FlowModule.cpp
@@ -2,9 +2,7 @@
 
 #include "FlowModule.h"
 
-#include "Modules/ModuleManager.h"
-
-#define LOCTEXT_NAMESPACE "Flow"
+#define LOCTEXT_NAMESPACE "FlowModule"
 
 void FFlowModule::StartupModule()
 {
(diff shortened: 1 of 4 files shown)


MaksymKapelianovych:
There is still one minor issue, which I decided to not include in this PR. If user remove/rename default pins "In/Out" in Params, node will still have them, because they are added in parent class constructor. This can lead to some troubles when user debugs graph, with question "why this pin is not getting triggered?".
First solution, that comes in mind, is to add 
```
UFlowNode_CallOwnerFunction::UFlowNode_CallOwnerFunction(const FObjectInitializer& ObjectInitializer)
...
{
...
    InputPins = {}; 
    OutputPins = {}; 
}
```
but then the node will appear in graph completely without pins and I don't know if it is OK.  Second solution is to made these pins unabled to remove from Params.

MothDoctor:
Thanks you for fixes and detailed explanation! :)

Issue #187, opened 2024-02-13
Nodes with Context Pins, loose ALL pins on CTRL+Z / can lead to crash


soraphis:
Reproduction:

- Create "Play Level Sequence" Node
- Assign any "LevelSequence"
- press CTRL+Z

Result:

![image](https://github.com/MothCocoon/FlowGraph/assets/3066582/4b28580c-fd1f-4006-8add-a8741af5c111)

My guess is that there is no "OnReconstructionRequested.ExecuteIfBound()" call for Undo!?

### To make it crash the editor:

Reproduction:

- Create "Play Level Sequence" Node
- Create Sequence Node
- Bind Level Sequence Outputs to Sequence Node input
- Assign any "LevelSequence"
- press CTRL+Z
- press CTRL+Y --> Crash

MothDoctor:
Thanks for a precise repro :)

I kinda managed to fix the first problem, pins won't disappear if I would call pin reconstruction from `PostEditUndo`. Sadly, it's not enough to fix a crash. More time is needed to dig into the scheme of transactions here.

This is what I have locally. 

``` cpp
void UFlowNode::PostEditUndo()
{
	UObject::PostEditUndo();

	// rebuild context pins on Undo any of node properties
	if (SupportsContextPins())
	{
		OnReconstructionRequested.ExecuteIfBound();
	}
}
```

MothDoctor:
Fixed by a pull request :)
https://github.com/MothCocoon/FlowGraph/pull/189

PR #189, opened 2024-03-24
Fix issue https://github.com/MothCocoon/FlowGraph/issues/187

MaksymKapelianovych:


MaksymKapelianovych:
The reason of that issue was, that during transactions, resetting pin arrays was not stored as part of the transaction, and after undo/redo action these arrays were empty and pins connections could not be properly restored or removed.
```
void UFlowGraphNode::ReconstructNode()
{
        ... 
	// Reset pin arrays
	Pins.Reset();
	InputPins.Reset();
	OutputPins.Reset();
         ...
}
```
To fix this we simply need to call Modify() before changing pins arrays. I decided to add this into the OnExternalChange() function and not into the ReconstructNode() itself, because in other places, where ReconstructNode() is called, transactions (and Modify() function call) exist in calling functions. 
I did not add transaction object itself, because if the OnExternalChange() is called from property modification event, it is already inside transaction, and if it is called from BP compilation, then this transaction would be stored as a separate transaction, and undoing it would not bring much use (to undo actual changes, user needs to press Ctrl+Z again).

MaksymKapelianovych:
This issue was reproduced with nodes PlayLevelSequence, CallOwnerFunction, SubGraph (when adding custom input/output), basically every node, that call OnReconstructionRequested.ExecuteIfBound() from its PostEditChangeProperty(). With my fix everything is working properly.

MothDoctor:
Thank you, it works beautifully! :)

PR #190, opened 2024-03-24
Improvements for working with Input/Output pins in BP FlowNodes

MaksymKapelianovych:
1. UFlowNode now also executes OnReconstructionRequested if pin info has been changed inside pin array (PinName, PinFriendlyName, PinToolTip).
2. Removed OnBlueprintPreCompile() and OnBlueprintCompiled() from UFlowGraphNode, because all possible node reconstructions, handled by these functions, now are triggered from UFlowNode::PostEditChangeProperty()
3. SFlowGraphNode now hides pin name only if there is one valid pin in array, not just Pins.Num() == 1.

### DIFF
--- a/Source/Flow/Private/Nodes/FlowNode.cpp
+++ b/Source/Flow/Private/Nodes/FlowNode.cpp
@@ -63,7 +63,8 @@ void UFlowNode::PostEditChangeProperty(FPropertyChangedEvent& PropertyChangedEve
 	Super::PostEditChangeProperty(PropertyChangedEvent);
 
 	if (PropertyChangedEvent.Property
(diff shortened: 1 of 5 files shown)


MaksymKapelianovych:
Before this change, FlowGraph reflected changes in BP FlowNode only after pressing Compile. If user wanted to revert changes, Ctrl+Z was reverting only changes in BP FlowNode, but not in the FlowGraph. To show changes in the Graph, Compile needed to be pressed again. 
Now every change inside Input/Output pins is immediately reflected in opened FlowGraphs, and pressing Ctrl+Z will revert changes both in BP FlowNode and FlowGraph (for proper usage this feature requires https://github.com/MothCocoon/FlowGraph/pull/189 to be merged)

PR #191, opened 2024-03-27
Fix crash in UFlowNode_CallOwnerFunction 

MaksymKapelianovych:
Add RequiredParamsClass validation to prevent crash in UFlowNode_CallOwnerFunction (with valid non-empty function ref) when user change owner class to default and then press Refresh


MaksymKapelianovych:
I decided not to set Params to nullptr, because we still have non-empty function name, and it would require to clear it as well. If user decides to set valid owner class back, function info for node will be lost, and user will need to set it back manually. Undo can help to revert changes, but there is a chance that some other nodes will be already modified until user notice and reverting all that work just to restore function info is not very convenient.

MaksymKapelianovych:
Steps to reproduce:
1. Set Expected Owner Class to some class, that contains function with correct signature for UFlowNode_CallOwnerFunction (let's call it Function1)
2. Add UFlowNode_CallOwnerFunction to graph and set its parameter Function to Function1
3. Change Expected Owner Class to some class, that does not contain Function1
4. Click Refresh

MothDoctor:
Accepted, thanks for this! :)

PR #192, opened 2024-04-01
Add graph action to delete node and reconnect pins if possible

MaksymKapelianovych:


MaksymKapelianovych:
Made this PR to make deleting simple nodes and reconnecting pins of adjacent nodes easier. I use Shift+Delete in Blueprints pretty often, especially when printing some debug staff and it is very practical. I am aware of SetPassThrough mode, but I think it would be useful to have this way also.
One difference from Blueprints is that I decided to allow pins reconnecting even if node has more than one input/output pins, but only first one should be connected. Usually user want to connect  first pins anyway, but if anything else is connected we can't predict how pins should be connected. If this change causes some inconveniences, I can modify PR, so only nodes with exactly one input and one output pins will utilize it.

MothDoctor:
Accepted, thanks for this! :)

PR #195, opened 2024-04-15
Fix missing UProperty

ymartel06:
Hi,

I've discovered few missing UProperty.
I've put UProperty as transient to keep same behaviors. They will not be saved or loaded. Properties tagged this way will be zero-filled at load time.
https://dev.epicgames.com/documentation/en-us/unreal-engine/unreal-object-handling-in-unreal-engine

Cheers,
-Yohann


MaksymKapelianovych:
Hi, all the structs you placed UPROPERTY macro in are not USTRUCTs, so, unfortunately, this will have no effect

PR #196, opened 2024-04-25
Fix LevelSequence Crash

youseiy:
we should set these after FinishSpawning is called


MaksymKapelianovych:
Hello. Quickly checked your PR, and I have small remark. There is the reason why those lines are before FinishSpawning - to avoid unnecesary replication after spawning actor. Whatever you do with replicated properties of an actor in between SpawnActorDeferred and FinishSpawning are replicated with initial spawning on clients. 
In engine version of similar function same approach is used and in my opinion it is better to keep it as it was, and fix crash for engine version prior to 5.4 by adding macro to check current engine version and substitute correct code snippet.

MothDoctor:
Yup, Maksym approach is correct one here :)

Issue #2, opened 2021-03-24
Duplicate FlowAsset Problem


yoyocyhsieh:
Hi,
When duplicate a FlowAsset in Content Browser, the NodeGuid of FlowGraphNode in duplicated FlowAsset will updated, but the NodeGuid of FlowNode remains the same, and UFlowAsset::TriggerInput function will failed because can't find the FlowNode.

PR #199, opened 2024-05-06
Flow Traits are now stored locally for every user

MaksymKapelianovych:
Breakpoints state (node and pins) was not serialized and after launching editor all breakpoints was disabled.


MothDoctor:
Yes, this is intended behavior. It would serialize the `Enabled` property, it would make it possible to submit assets with enabled breakpoints to the repository. That would later infuriate other team members - suspending the game out of the blue. So they would disable breakpoint and submit to the repository. It would be a vicious cycle.

That's why blueprints only save "a breakpoint was placed here", but not an "Enabled" state.

Thus I cannot accept change in this form. I would gladly take it in, if you would find a way to keep this state as local, user-specific data. Not as information serialized to the game asset itself.

MaksymKapelianovych:
Thanks for your quick response! 
Yes, very good argument for not making this serializable. I will try to find the way how it is done in Blueprints and implement this properly.

MaksymKapelianovych:
Moved traits from UFlowGraphNode to UFlowGraphEditorSettings, which is using config EditorPerProjectUserSettings, so now traits will not be part of an asset. I decided to store them per node, as it was previously. Also enum EFlowTraitType was added to leave possibility to easily add new trait in future. Previously it was enough to add new UFlowGraphNode member `FFlowPinTrait Something;`, but now it is not the case anymore, so I decided enum would be optimal solution. All functions to work with traits now contained in UFlowDebuggerSubsystem as static functions, FFlowDebugTrait can be modified only by UFlowDebuggerSubsystem.
For reference was taken UBlueprintEditorSettings and FKismetDebugUtilities.

Tested scenarios:
1. Add/remove/toggle/disable breakpoints and restarting editor - everything restores as it should
2. Remove all breakpoints - PerNodeSettings map is cleared in config
3. Delete pin/node with added breakpoint - breakpoint is deleted from PerNodeSettings
4. Rename pin (regular and context pins) - new pin is generated and new one is created, so breakpoint is deleted with the old one
5. Breakpoint actions on ghost nodes (with unset or deleted UFlowNode) - everything work as for normal node
6. Breakpoins are not duplicated when duplicating node/graph

Source/Flow/Public/Nodes/FlowPin.h
```diff
@@ -210,48 +210,3 @@ struct FLOW_API FPinRecord
 	FORCEINLINE static FString DoubleDigit(const int32 Number);
 };
 #endif
-
-// It can represent any trait added on the specific node instance, i.e. breakpoint
-USTRUCT()
-struct FLOW_API FFlowPinTrait
```
MaksymKapelianovych, line 408: Moved FFlowPinTrait to UFlowDebuggerSubsystem as it is used only in Editor module and represents only debug feature

Source/FlowEditor/Public/Asset/FlowDebuggerSubsystem.h
```diff
@@ -6,9 +6,77 @@
...
+
+	// ^ Add new trait types above here ^
+	Max
+};
+
+ENUM_RANGE_BY_COUNT(EFlowTraitType, EFlowTraitType::Max)
+
+// It can represent any trait added on the specific node instance, i.e. breakpoint
+USTRUCT()
+struct FLOWEDITOR_API FFlowDebugTrait
```
MaksymKapelianovych: Renamed FFlowPinTrait to FFlowDebugTrait to better show purpose of this struct. Can rename back if needed)

MaksymKapelianovych:
It is very easy to add new trait types. For example, we want to add "Log" trait, which will print to log node/pin execution.
First we need to add new type to the EFlowTraitType:
image_2024-05-14_16-13-11

then we need to define how this trait will work:
image_2024-05-14_16-14-52
image_2024-05-14_16-14-52 (2)
image_2024-05-14_16-14-52 (3)

and create and bind all corresponding UI actions

And after adding Log trait to some nodes and pins we get this result:
image_2024-05-14_16-14-54

Of course new trait will require some visual representation and organizing its actions in context menu, but I think it is not urgent right now.

Source/FlowEditor/Public/Graph/FlowGraphEditorSettings.h
```diff
@@ -13,6 +14,24 @@ enum class EFlowNodeDoubleClickTarget : uint8
...
+USTRUCT()
+struct FLOWEDITOR_API FFlowTraitSettings
+{
+	GENERATED_BODY()
+	
+	UPROPERTY()
+	TArray<FFlowDebugTrait> NodeTraits;
+
+	UPROPERTY()
+	TArray<FFlowDebugTrait> PinTraits;
```
MaksymKapelianovych: Store as TArray and not as TMap<FGuid, TArray< FFlowDebugTrait >> (where Key is Guid of Pin), because usually there will be very small amount of traits, and I think TArray would be better choice as all its elements are small and grouped in the same memory, so it is more cache-friendly.

Source/FlowEditor/Public/Asset/FlowDebuggerSubsystem.h
```diff
@@ -34,4 +102,85 @@ class FLOWEDITOR_API UFlowDebuggerSubsystem : public UEditorSubsystem
...
+	/** Returns the enabled flag for the OwnerPin's trait with provided Type */
+	static bool IsTraitEnabled(const UEdGraphPin* OwnerPin, EFlowTraitType Type);
+
+	/** Creates or removes OwnerNode's trait with provided Type */
+	static void ToggleTrait(const UFlowGraphNode* OwnerNode, EFlowTraitType Type);
+	/** Creates or removes OwnerPin's trait with provided Type */
+	static void ToggleTrait(const UEdGraphPin* OwnerPin, EFlowTraitType Type);
+
+	/** Sets the hit flag for the all OwnerNode's traits */
+	static TArray<EFlowTraitType> SetAllTraitsHit(const UFlowGraphNode* OwnerNode, bool bHit);
```
MaksymKapelianovych: Maybe it would be better to return bitmask here? I doubt that in future there will be over 32/64 trait types

MothDoctor:

MaksymKapelianovych:
Updated!

MothDoctor:
Hey, I apologize I'm coming back after so long. I needed time to think about potential features that users might want to add in future. I'm willing to accept this PR, but I see a need for a bit of refactoring.

Moving all things to the editor module makes sense when looking at the current codebase. 
- But what if someone would like to build a Flow Graph representation as UI overall available in a cooked game? This might be a read-only graph, used for debugging purposes. Especially useful in multiplayer, and on consoles.
- Even without the feature described above, I would push things forward a better debugging in multiplayer. I admit that currently, it works well only for debugging server-side Flow Graph instances in PIE.
- What you did already brings us closer to addressing the above concerns. It takes out breakpoint data out of the Flow Asset.

From this perspective, I'd like to propose the following changes.
- Creating a new `Developer` module where we would move all your changes to `UFlowDebuggerSubystem` which would need to extend `UWorldSubsystem`. It will work as the engine's UEdGraph types are placed in the runtime module. However, FlowEditor types like `UFlowGraphNode` can't be included there.
- The existing subsystem should be renamed `UFlowDebugEditorSubsystem` which I will promptly do.
- None of the methods or data there can be `static`, as this wouldn't work well while testing the game in Listen Server PIE-based configuration (multiple world instances as part of the same process). For purposes of integrating this changelist, I would only expect that the breakpoint set in a graph of a given server/client instance, is active only for this specific instance. Anything more sophisticated might be non-trivial to implement (how user would control if pin breakpoint is set only for this instance or all instances? how to support all networking scenarios?)
- I would move `FFlowPinTrait` to this Developer module. You're right, it doesn't seem useful in the Runtime module. I would keep the old struct name, however. Any project might reuse this structure in different unforeseen ways. And needs to be moved to its own header, other classes shouldn't include the Debugger Subsystem class.
- I would remove `EFlowTraitType` as this would require adding a new enum element for any custom need. Later on, 
- I would move `FFlowTraitSettings` to the same headers as `FFlowPintTrait`. I'd add there Flow Asset guide as well and rename it `FFlowAssetTrait`.
- Surely, trait records cannot live in `UFlowGraphEditorSettings`, as this is where set project-wide settings are. We need to move anyway to a new class in the Developer module. 
- If anyone would like to create a new trait type, I would assume they would reuse existing data types and functions. Although they would need to add their property to hold their `FFlowAssetTrait`.

What do you think? :)
I promise to conduct follow-up reviews much quicker.

PS I also encountered a weird bug while doing the initial test of the current code. It looks like only the first node breakpoint is respected. The debugger doesn't "hit" the second one...
image

MaksymKapelianovych:
About the bug: it exists in previous version of "trait" system too. I didn't change the behaviour of traits, just refactored it to be able to store trait's state between editor sessions and to not affect other devs. Regarding any known and unknown bugs - this PR does not fix them.

Why does this bug exist: breakpoints in Flow stop only editor, C++ code continues to run. In your example player is already loaded when OnActorRegistered activates, and this node immidiately triggers Success pin, passing the signal to Sequence node. Editor session is being requested to pause, yes, but C++ functions does not know about it and Flow keeps activating nodes until it hits "wait" nodes, that pause the signal. You can see it yourself if you launch your example one more time and look at the graph. Both of your breakpoint will be rendered as "hit". Breakpoints in any subsequent node after Sequence node will also be rendered as "hit".
image


To fix this, we have to introduce some pausing system directly into the UFlowAsset, which will trigger subsequent nodes only if play session is not paused. And after play is resumed it should resume all waiting signals (should be only one).

MaksymKapelianovych:
> * But what if someone would like to build a Flow Graph representation as UI overall available in a cooked game? This might be a read-only graph, used for debugging purposes. Especially useful in multiplayer, and on consoles.

  For users to be able to this we surely have to fix mentioned bug) But I am not sure how pausing will work in cooked game, cause now it requires to invoke UEditorEngine::PlaySessionPaused and obviously it can be called in cooked game. Not sure if it is needed though. 
All in all I want to say that for this scenario I can't guarantee it will work at all, and I can't properly test it, because it would require to build some test UI by myself.

> * Even without the feature described above, I would push things forward a better debugging in multiplayer. I admit that currently, it works well only for debugging server-side Flow Graph instances in PIE.

I can add server/client type selection for breakpoins. And add a check in debugger if owner role is matching breakpoint type. This shouldn't be hard to do.

MaksymKapelianovych:
> * (how user would control if pin breakpoint is set only for this instance or all instances? how to support all networking scenarios?)
Regarding these questions I think we can at least make it similar to BP's behaviour: if some instance is selected as inspected, then breakpoints are only triggered for this specific instance, ignoring all other instances. For networking we can do similar thing and select wanted world for instance.
image

MaksymKapelianovych:
> * I would move `FFlowPinTrait` to this Developer module. You're right, it doesn't seem useful in the Runtime module. I would keep the old struct name, however.

I renamed it, because it is not pin-only, nodes can have traits too. For me, name FFlowPinTrait is a little bit confusing, at first I thought it can only be applied to pins and searched for FFlowNodeTrait)

MaksymKapelianovych:
> * If anyone would like to create a new trait type, I would assume they would reuse existing data types and functions. Although they would need to add their property to hold their `FFlowAssetTrait`.

I don't quite understand how adding new trait types would look like. Will it require for users to add the whole set of enable/disable/add/remove/etc functions for every new type?
I think it is much easier to add a new enum element, add required buttons to context menu and add a processing block into UFlowGraphNode::TryPausingSession (it probably needs a rename).

Maybe I don't fully understand what do you mean by "trait" and what possible types/behaviours there can be? I see it only as a debugging thing.

MothDoctor:
> > * I would move `FFlowPinTrait` to this Developer module. You're right, it doesn't seem useful in the Runtime module. I would keep the old struct name, however.
> 
> I renamed it, because it is not pin-only, nodes can have traits too. For me, name FFlowPinTrait is a little bit confusing, at first I thought it can only be applied to pins and searched for FFlowNodeTrait)

OK, this convinces me. Let us keep your new name :)

> > * But what if someone would like to build a Flow Graph representation as UI overall available in a cooked game? This might be a read-only graph, used for debugging purposes. Especially useful in multiplayer, and on consoles.
> 
> For users to be able to this we surely have to fix mentioned bug) But I am not sure how pausing will work in cooked game, cause now it requires to invoke UEditorEngine::PlaySessionPaused and obviously it can be called in cooked game. Not sure if it is needed though. All in all I want to say that for this scenario I can't guarantee it will work at all, and I can't properly test it, because it would require to build some test UI by myself.

Yeah, we would need to implement a different way of pausing the game session. But this is out of the scope of this PR.
To push forward this acceptance of this PR, I took the freedom to move your new subsystem code into a separate class in the new `DeveloperTool` module.
Now it will be possible to resolve conflict with the renamed old class of editor-only subsystem.

`PausePlaySession` is now an empty virtual method in the new class. I plan to inherit the old editor-only subsystem after the new class.

> > * Even without the feature described above, I would push things forward a better debugging in multiplayer. I admit that currently, it works well only for debugging server-side Flow Graph instances in PIE.
> 
> I can add server/client type selection for breakpoins. And add a check in debugger if owner role is matching breakpoint type. This shouldn't be hard to do.

Awesome! To prepare for multiplayer and developer/editor split, I already changed all methods in the subsystem from `static` to `virtual`.

MaksymKapelianovych:
Why the methods have to be virtual though and not static? 🤔 I don't quite understand intended purpose

P.S. Good to finally see the FFlowBreakpoint, much better now)

MothDoctor:
Now, in the next commit, I was able to make the new `UFlowDebuggerSubsystem` a parent to `UFlowDebugEditorSubsystem`.

Additionally
I cleaned up most comments above Trait methods, as method names are self-explanatory. It's easier to read code without rudimentary comments. I also renamed some methods for even better readability.
- I removed `EFlowTraitType` entirely, as I don't see use cases for other traits here. This addition made breakpoint logic relatively hard to read. If somebody would like to create some node/pin metadata, new data should be stored separately.
- Remaining struct types have been moved to `FlowDebuggerTypes.h`.
- In the end, I renamed `FFlowDebugTrait` to `FFlowBreakpoint` as this finally conveys its purpose. This is followed by renaming methods and rest code to use the word `breakpoint` instead of `trait`.
- Restored distinction between `MarkAsHit` and `ResetHit` methods as micro-optimization. It is also easier to read.
- Eliminated `check` as we truly do not want to crash the game due to some stale data, it might prevent the user from debugging the actual issue ;)
- Eliminated `checkSlow` as well.

After all was done, I realized that keeping together Node and Pin breakpoints (especially in an array) does cost us some cycles on every single Input and Output pin triggered. In this case, the code needs to access Node, find its entry, and then iterate the array to find the pin. It's a small operation, but triggered so frequently that might cost us more (with many breakpoints) than cost of accessing memory stored in the hash map. In effect, I did refactored that structure to 2 TMaps. It's fast and simplifies code again.

MaksymKapelianovych:
> After all was done, I realized that keeping together Node and Pin breakpoints (especially in an array) does cost us some cycles on every single Input and Output pin triggered. In this case, the code needs to access Node, find its entry, and then iterate the array to find the pin. It's a small operation, but triggered so frequently that might cost us more (with many breakpoints) than cost of accessing memory stored in the hash map. In effect, I did refactored that structure to 2 TMaps. It's fast and simplifies code again.

Nice, I was thinking about it too, ~~but was not sure about if that was intended~~ nevermind)

MothDoctor:
> Why the methods have to be virtual though and not static? 🤔 I don't quite understand intended purpose
> 
> P.S. Good to finally see the FFlowBreakpoint, much better now)

With the `static` method the same code is called for every instance of the multiplayer world, in case we're testing in a single process. Although I might over-engineered code by making that `virtual`, assuming that project-specific code might want to add their events. But now I'm thinking "wait, both Slate and Dear ImGUI implementation pull data every single frame" ;) 

However, we will need a way to distinguish players, potentially networked sessions and local split-screen (with different `ULocalPlayer`). Here is where a non-static object can pull relevant data (which world is that, so is that server or client #2, which Local Player is used) as this shouldn't be provided by `UFlowGraphNode`.

> > After all was done, I realized that keeping together Node and Pin breakpoints (especially in an array) does cost us some cycles on every single Input and Output pin triggered. In this case, the code needs to access Node, find its entry, and then iterate the array to find the pin. It's a small operation, but triggered so frequently that might cost us more (with many breakpoints) than cost of accessing memory stored in the hash map. In effect, I did refactored that structure to 2 TMaps. It's fast and simplifies code again.
> 
> Nice, I was thinking about it too, but was not sure about if that was intended

Oh, I forgot to mention. There's one caveat. For now, we don't clear breakpoints for pins removed during node reconstruction, as I couldn't iterate all pin breakpoints of a single node. Which isn't a noticeable problem, although leaving useless data in .ini. One way to tackle that would be adding a "remove all breakpoints" action to the Flow Asset editor which is a new power we gained thanks to your approach. Now it would be trivial to clear all Flow Graph breakpoints, like we do in coding IDE :)

MaksymKapelianovych:
> However, we will need a way to distinguish players, potentially networked sessions and local split-screen (with different ULocalPlayer). Here is where a non-static object can pull relevant data (which world is that, so is that server or client https://github.com/MothCocoon/FlowGraph/issues/2, which Local Player is used) as this shouldn't be provided by UFlowGraphNode.

So, if I understand it right, you are implying possibility to add "per-world" breakpoints? Or "per-player"? But there still be a single source of breakpoints, .ini file, without distinction between worlds/etc. So every world will have the same set of breakpoints and we clearly will need a different way to handle described cases. For now I can see it only for client/server.

> For now, we don't clear breakpoints for pins removed during node reconstruction, as I couldn't iterate all pin breakpoints of a single node. Which isn't a noticeable problem, although leaving useless data in .ini.

Hm hm, I remember there was a method called `CleanupTraits`, which does exactly this, not right away, but when breakpoint list is updated. Is it not working?

MothDoctor:
> So, if I understand it right, you are implying possibility to add "per-world" breakpoints? Or "per-player"?

Exactly, per-world breakpoints in the multiplayer game if a single process hosts multiple worlds. One world per player, but only one process. It's not an issue in the packaged, networked game (separate game process for every person playing). But it needs to be supported for developer tests. That's why using anything static is no-go in multi.

And split-screen, most likely. I'm pretty sure someone will use separate Flow Asset instances per player ;) 

> Hm hm, I remember there was a method called `CleanupTraits`, which does exactly this, not right away, but when breakpoint list is updated. Is it not working?

It's still kinda there as `UFlowDebuggerSubsystem::RemoveObsoletePinBreakpoints`, but it's not that trivial to handle this case when Node and Pin breakpoints are separated. This can be done, I just wanted first to integrate this change with Homer's optimization. We got now solid foundation for a way more advanced debugger :) 

image

MaksymKapelianovych:
> > So, if I understand it right, you are implying possibility to add "per-world" breakpoints? Or "per-player"?
> 
> Exactly, per-world breakpoints in the multiplayer game if a single process hosts multiple worlds. One world per player, but only one process. It's not an issue in the packaged, networked game (separate game process for every person playing). But it needs to be supported for developer tests. That's why using anything static is no-go in multi.
> 
> And split-screen, most likely. I'm pretty sure someone will use separate Flow Asset instances per player ;)

Very excited to see how people would do it for split screen)
And with per-world multiplayer breapoints, just to clarify, is it just "server", "client 1/2/3/4"?

MaksymKapelianovych:
If it is just "specify client/player for which this breakpoint will be set" I think we can add context menu to breakpoints, where user can select/deselect all needed "worlds". I will show tomorrow (I'll try) how this context menu would look like. And it would we good to show only needed options (for example if it is co-op for 2 players there will be no need to show "client 3/4", I think we can extract number of clients from play settings)

MothDoctor:
> If it is just "specify client/player for which this breakpoint will be set" I think we can add context menu to breakpoints, where user can select/deselect all needed "worlds". I will show tomorrow (I'll try) how this context menu would look like. And it would we good to show only needed options (for example if it is co-op for 2 players there will be no need to show "client 3/4", I think we can extract number of clients from play settings)

Well, have fun! :) 

In meantime I've done this refactor in 1a4a2d95b441108200a54e4d077af9d08fe93e12
> Refactored old breakpoint logic to open doors to building a separate cook-only graph debugger. If anyone would be willing to implement such a feature, of course. This would require much work, but the Flow Graph community keeps surprising!
        * Decoupled triggering breakpoints from the Flow Editor module. `UFlowGraphNode` no longer operates on breakpoints. Instead, `UFlowDebuggerSubsystem` binds directly to the runtime `OnPinTriggered` delegate.
        * Pin breakpoint is now identified as `NodeGuid` and `PinName` instead of `UEdGraphPin`. Thanks to this it's now possible to bind to `OnPinTriggered` delegate outside of the editor! Events shall be received in the Standalone game, non-shipping game builds.
        * Wrapped `UEdGraphNod` pointer in runtime Flow Node class with `WITH_EDITORONLY_DATA` since this isn't even loaded in a Standalone game.
        * Now it's possible to add custom logic to "On Pin Triggered" logic outside of the runtime module. You can add what you want by extending UFlowDebuggerSubsystem or UFlowDebugEditorSubsystem.
        
This work accidentally solved the regression I caused previously. I had to return your struct (node breakpoint and pin breakpoints as one struct), so I could identify pin breakpoints by `NodGuid` and `PinName`. Thanks to this I could restore clearing breakpoints after pins were deleted during node reconstruction.

MaksymKapelianovych:
So, I have two options for it:
The first one is how Unreal handles debugging BP in multiplayer. I prefer this one, because users are already familiar with it and it will be easy to implement.
image

The second one is to setup breakpoint type directly within breakpoint itself. Given types list is debatable, it can be as on screen or as in EFlowNetMode.
image
image

If you have some preferences or considerations regarding this, I would love to read it. Otherwise, I will stick to the first option)

MothDoctor:
Yeah, option 1 lets users choose which networked instance should be observed. Otherwise, it's easy to confuse which breakpoint observes what. Let's go with Option 1 and continue the conversation in the new PR :)

PR #202, opened 2024-05-09
Flow AddOns

LindyHopperGT:
Introduced FlowAddOns

~ Major rework of the Flow Editor to support AddOns ~ Also added ExecuteComponent flow node, which allows a component of the owning actor (if it implements a specific interface) to be executed via a proxy node in the Flow Graph.  We use this to wrap some legacy systems in flow graph nodes.


Config/BaseFlow.ini
```diff
@@ -1,5 +1,6 @@
 [CoreRedirects]
 +ClassRedirects=(OldName="/Script/Flow.FlowNode_CustomEvent",NewName="/Script/Flow.FlowNode_CustomInput")
 +PropertyRedirects=(OldName="FlowAsset.CustomEvents",NewName="CustomInputs")
++PropertyRedirects=(OldName="FlowGraphNode.FlowNode",NewName="NodeInstance")
```
LindyHopperGT, line 4: Renamed this member to make the conversion for FlowGraphNode operating on FlowNodeBase easier and clearer, when casting to FlowNodes.

It also has a precedent in a lot of the UE UEdGraphNodes

Source/Flow/Private/AddOns/FlowNodeAddOn.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+void UFlowNodeAddOn::Finish()
+{
+	if (ensure(FlowNode))
+	{
+		FlowNode->Finish();
+	}
+}
+
+EFlowAddOnAcceptResult UFlowNodeAddOn::AcceptFlowNodeAddOnParent_Implementation(const UFlowNodeBase* ParentTemplate) const
```
LindyHopperGT, line 48: This is the function on the AddOn-side to make itself eligible to a parent Node or AddOn that knows nothing about it. 

This case is useful for AddOns in plugins to well-established FlowNodes, where we don't want to modify the Node to add eligibility checks for the new AddOn.

Source/Flow/Private/AddOns/FlowNodeAddOn.cpp
```diff
@@ -0,0 +1,78 @@
...
+	// during runtime and that we are not calling this method before the addon has been
+	// initialized.
+	ensure(FlowNode);
+
+	return FlowNode;
+}
+
+void UFlowNodeAddOn::CacheFlowNode()
+{
+	UObject* OuterObject = GetOuter();
```
LindyHopperGT, line 84: At runtime, the AddOn caches a pointer to its owning FlowNode to optimize lookups (since it may be multiple Outer's away from the AddOn itself).

Source/Flow/Private/AddOns/FlowNodeAddOn_PredicateAND.cpp
```diff
@@ -0,0 +1,53 @@
...
+{
+#if WITH_EDITOR
+	NodeStyle = EFlowNodeStyle::Logic;
+	Category = TEXT("Composite");
+#endif
+}
+
+EFlowAddOnAcceptResult UFlowNodeAddOn_PredicateAND::AcceptFlowNodeAddOnChild_Implementation(const UFlowNodeAddOn* AddOnTemplate) const
+{
+	if (IFlowPredicateInterface::ImplementsInterfaceSafe(AddOnTemplate))
```
LindyHopperGT, line 18: Here we see the AddOn deciding eligibility for its potential child AddOns, which is the opposite direction from the Parent-side version mentioned in a previous comment above.

In this case, the AddOn or Node parent knows what child AddOns it supports and filters them out with this function.

Source/Flow/Private/FlowAsset.cpp
```diff
@@ -139,25 +134,25 @@ EDataValidationResult UFlowAsset::ValidateAsset(FFlowMessageLog& MessageLog)
 	return MessageLog.Messages.Num() > 0 ? EDataValidationResult::Invalid : EDataValidationResult::Valid;
 }
 
-bool UFlowAsset::IsNodeClassAllowed(const UClass* FlowNodeClass, FText* OutOptionalFailureReason) const
+bool UFlowAsset::IsNodeOrAddOnClassAllowed(const UClass* FlowNodeOrAddOnClass, FText* OutOptionalFailureReason) const
```
LindyHopperGT, line 139: This function (and a number of others like it) expanded to support Nodes and AddOns, not just Nodes.

Source/Flow/Public/Interfaces/FlowContextPinSupplierInterface.h
```diff
@@ -0,0 +1,45 @@
...
+// A flow element (UFlowNode, UFlowNodeAddOn, etc.) that may supply context pins
+// "Context Pins" are those that can be dynamically added/removed to a FlowNode by property
+// settings on the flow node, by subobjects, etc.
+UINTERFACE(MinimalAPI, Blueprintable, DisplayName = "Flow ContextPin Supplier Interface")
+class UFlowContextPinSupplierInterface : public UInterface
+{
+	GENERATED_BODY()
+};
+
+class FLOW_API IFlowContextPinSupplierInterface
```
LindyHopperGT, line 19: This is an interface for the ContextPin supplying functions (subgraph nodes, etc.)

Source/Flow/Public/Interfaces/FlowCoreExecutableInterface.h
```diff
@@ -0,0 +1,63 @@
...
+
+// Implemented by objects that can execute within a flow graph
+// (eg, UFlowNode and UFlowNodeAddOn subclasses implement this)
+UINTERFACE(MinimalAPI, Blueprintable, DisplayName = "Flow Core Executable Interface")
+class UFlowCoreExecutableInterface : public UInterface
+{
+	GENERATED_BODY()
+};
+
+class FLOW_API IFlowCoreExecutableInterface
```
LindyHopperGT, line 17: This is the interface for the "Core" Flow node & add-on functions.  

These are core to the experience of an executable element of a flow graph.

Source/Flow/Public/Interfaces/FlowExternalExecutableInterface.h
```diff
@@ -0,0 +1,33 @@
...
+
+// Implemented by external objects that can execute within a flow graph
+// via a FlowNode or FlowNodeAddOn proxy
+UINTERFACE(MinimalAPI, Blueprintable, DisplayName = "Flow External Executable Interface")
+class UFlowExternalExecutableInterface : public UInterface
+{
+	GENERATED_BODY()
+};
+
+class FLOW_API IFlowExternalExecutableInterface
```
LindyHopperGT, line 21: This interface can be implemented by "external" (ie, non-flow node/addon) objects to become executable via a proxy node or addon in a flow graph.

See the ExecuteComponent flow node for example usage.  If the component implements this interface and the Core flow interface, it can be executed by that node.

Source/Flow/Public/Interfaces/FlowNativeExecutableInterface.h
```diff
@@ -0,0 +1,44 @@
...
+
+// Implemented only by native (C++) objects that can execute within a flow graph
+// (eg, UFlowNode and UFlowNodeAddOn subclasses implement this)
+UINTERFACE(MinimalAPI, BlueprintType, DisplayName = "Flow Native Executable Interface", meta = (CannotImplementInterfaceInBlueprint))
+class UFlowNativeExecutableInterface : public UInterface
+{
+	GENERATED_BODY()
+};
+
+class FLOW_API IFlowNativeExecutableInterface
```
LindyHopperGT, line 18: These are the flow executable functions that ONLY native flow classes UFlowNode and UFlowNodeAddOn implement.

They bridge the node/addon to the rest of the graph (triggering, and finishing).

Source/Flow/Public/Interfaces/FlowPredicateInterface.h
```diff
@@ -0,0 +1,30 @@
...
+class UFlowNodeAddOn;
+
+// Predicate interface for AddOns
+UINTERFACE(MinimalAPI, BlueprintType, Blueprintable, DisplayName = "Flow Predicate Interface")
+class UFlowPredicateInterface : public UInterface
+{
+	GENERATED_BODY()
+};
+
+class FLOW_API IFlowPredicateInterface
```
LindyHopperGT, line 19: This is an interface for Flow "Predicate" AddOns that can be used by the "Branch" flow node.

Source/Flow/Public/Nodes/FlowNode.h
```diff
@@ -50,30 +45,12 @@ class FLOW_API UFlowNode : public UObject, public IVisualLoggerDebugSnapshotInte
 
 	UPROPERTY()
 	TArray<TSubclassOf<UFlowAsset>> DeniedAssetClasses;
-
```
LindyHopperGT, line 53: Much of the FlowNode properties moved to FlowNodeBase, with the notable exceptions of those having to do with pins and focus.

Also, many of the functions moved to FlowNodeBase or into the Interfaces' implementations.

Source/Flow/Public/Nodes/FlowNode.h
```diff
@@ -82,52 +59,18 @@ class FLOW_API UFlowNode : public UObject, public IVisualLoggerDebugSnapshotInte
...
-	// Short summary of node's content - displayed over node as NodeInfoPopup
-	UFUNCTION(BlueprintImplementableEvent, Category = "FlowNode", meta = (DisplayName = "Get Node Description"))
-	FString K2_GetNodeDescription() const;
-
 	// Inherits Guid after graph node
 	UPROPERTY()
 	FGuid NodeGuid;
 
 public:
-	UFUNCTION(BlueprintCallable, Category = "FlowNode")
```
LindyHopperGT: Un-UFunctioning this was a merge mistake I think.  I think I might want to revert these changes.

Source/Flow/Public/Nodes/FlowPin.h
```diff
@@ -2,9 +2,10 @@
 
 #pragma once
 
+#include "UObject/ObjectMacros.h"
 #include "FlowPin.generated.h"
 
-USTRUCT()
+USTRUCT(BlueprintType)
```
LindyHopperGT, line 8: Some Blueprint FlowNodes/AddOns use this struct directly.

Source/Flow/Public/Nodes/World/FlowNode_CallOwnerFunction.h
```diff
@@ -24,60 +26,64 @@ class FLOW_API UFlowNode_CallOwnerFunction : public UFlowNode
...
+	bool TryAllocateParamsInstance();
+
 	// Helper function for DoesFunctionHaveValidFlowOwnerFunctionSignature()
 	static bool DoesFunctionHaveNameReturnType(const UFunction& Function);
-#endif // WITH_EDITOR
+
+protected:
+
+	// Function reference on the expected owner to call
+	UE_DEPRECATED(5.4, "Sunsetting this feature from FlowGraph with UE 5.4.  Custom FlowNodes are a better mechanism to use.");
```
LindyHopperGT: As discussed, marking the CallOwnerFunction feature as Deprecated and scheduled for demolition.

Source/FlowEditor/FlowEditor.Build.cs
```diff
@@ -12,7 +12,8 @@ public FlowEditor(ReadOnlyTargetRules target) : base(target)
 		{
 			"EditorSubsystem",
 			"Flow",
-			"MessageLog"
+			"MessageLog",
+			"AIModule", // For BlueprintNodeHelpers::DescribeProperty (could be copy/pasted out to remove editor-only dependency)
```
LindyHopperGT, line 16: This is an editor-only dependency on the AIModule, so I didn't think it was a problem.

But it could be removed by copying or otherwise reimplementing the DescribeProperty function that is used.

Source/Flow/Public/Nodes/FlowNode.h
```diff
@@ -82,52 +59,18 @@ class FLOW_API UFlowNode : public UObject, public IVisualLoggerDebugSnapshotInte
...
-	virtual FString GetNodeDescription() const;
 #endif
 
-protected:
-	// Short summary of node's content - displayed over node as NodeInfoPopup
-	UFUNCTION(BlueprintImplementableEvent, Category = "FlowNode", meta = (DisplayName = "Get Node Description"))
-	FString K2_GetNodeDescription() const;
-
 	// Inherits Guid after graph node
 	UPROPERTY()
```
LindyHopperGT, line 64: Only full UFlowNodes (not AddOns) are assigned GUIDs and are stored in the FlowAsset's node array.

This means they are also the only things saved in the save state code as well.

This may be something that changes in the future.

Source/FlowEditor/Private/Graph/FlowGraphEditor.cpp
```diff
@@ -561,25 +552,53 @@ bool SFlowGraphEditor::CanCutNodes() const
 
 void SFlowGraphEditor::CopySelectedNodes() const
 {
-	const FGraphPanelSelectionSet SelectedNodes = GetSelectedNodes();
-	for (FGraphPanelSelectionSet::TConstIterator SelectedIt(SelectedNodes); SelectedIt; ++SelectedIt)
+	// Export the selected nodes and place the text on the clipboard
```
LindyHopperGT, line 555: Copy and Paste underwent a major revision to handle copying and pasting a tree of subnodes.

Source/FlowEditor/Private/Graph/FlowGraphEditor.cpp
```diff
@@ -609,83 +628,179 @@ void SFlowGraphEditor::PasteNodes()
...
+	const bool bIsPastePossible = FEdGraphUtilities::CanImportNodesFromText(FlowAsset->GetGraph(), ClipboardContent);
 
-		return FEdGraphUtilities::CanImportNodesFromText(FlowAsset->GetGraph(), ClipboardContent);
+	if (!bIsPastePossible)
+	{
+		return false;
 	}
 
-	return false;
+	// TODO (gtaylor) Need to confirm the nodes are allowed to be pasted on the selected node(s)
```
LindyHopperGT, line 801: This is a known problem.  You can paste AddOns where they are not normally allowed, at this time.

We'll be fixing that later this month~.

Source/FlowEditor/Private/Graph/FlowGraphSchema_Actions.cpp
```diff
@@ -197,6 +200,69 @@ UFlowGraphNode* FFlowGraphSchemaAction_NewNode::ImportNode(UEdGraph* ParentGraph
 	return NewGraphNode;
 }
 
+/////////////////////////////////////////////////////
+// New SubNode (AddOn)
```
LindyHopperGT, line 204: "SubNode" is the editor-side word for "AddOn" (runtime)

Source/FlowEditor/Private/Nodes/FlowNodeBlueprintFactory.cpp
```diff
@@ -34,13 +38,14 @@ class SFlowNodeBlueprintCreateDialog final : public SCompoundWidget
 {
 public:
 	SLATE_BEGIN_ARGS(SFlowNodeBlueprintCreateDialog) {}
+		SLATE_ARGUMENT(TSubclassOf<UFlowNodeBase>, ParentClass)
```
LindyHopperGT, line 41: I generalized this factory so that it could make both of the Blueprint types (Node and AddOn) with thin subclasses, below.

Source/FlowEditor/Private/UnrealExtensions/IFlowCuratedNamePropertyCustomization.cpp
```diff
@@ -33,19 +36,24 @@ void IFlowCuratedNamePropertyCustomization::Initialize()
 	check(!CachedTextSelected.IsValid());
 	check(CachedTextList.IsEmpty());
 
-	const FName CuratedName = GetCuratedName();
-	const bool bChangedValue = TrySetCuratedNameWithSideEffects(CuratedName);
+	FName CuratedName;
```
LindyHopperGT, line 39: These are some cumulative bugfixes for the Curated Name List feature.

Source/FlowEditor/Public/Graph/Nodes/FlowGraphNode.h
```diff
@@ -202,4 +216,107 @@ class FLOWEDITOR_API UFlowGraphNode : public UEdGraphNode
 
 	virtual EFlowSignalMode GetSignalMode() const;
 	virtual bool CanSetSignalMode(const EFlowSignalMode Mode) const;
+
+//////////////////////////////////////////////////////////////////////////
```
LindyHopperGT, line 220: SubNode support grafted on from the EnvironmentQuery graph editor (mostly) as a guide, and then modified from there to blend it in with the existing FlowEditor.

Source/FlowEditor/Public/UnrealExtensions/IFlowExtendedPropertyTypeCustomization.h
```diff
@@ -63,7 +63,8 @@ StructT* IFlowExtendedPropertyTypeCustomization::TryGetTypedStructValue(const TS
 		TArray<void*> RawData;
 		StructPropertyHandle->AccessRawData(RawData);
 
-		if (RawData.Num() > 0)
+		// (must be exactly one, multi-select is not supported for this feature)
```
LindyHopperGT, line 66: multiselect was a big problem with the curated name list.  so we're just punting on it.  if you multiselect, it just doesn't show you the drop-down at all.

Source/Flow/Public/Nodes/FlowNode.h
```diff
@@ -125,9 +74,6 @@ class FLOW_API UFlowNode : public UObject, public IVisualLoggerDebugSnapshotInte
 	UFUNCTION(BlueprintPure, Category = "FlowNode")
 	const FGuid& GetGuid() const { return NodeGuid; }
 
-	UFUNCTION(BlueprintPure, Category = "FlowNode")
-	UFlowAsset* GetFlowAsset() const;
```
LindyHopperGT, line 129: This moved to UFlowNodeBase

Source/Flow/Public/Interfaces/FlowCoreExecutableInterface.h
MothDoctor: I have doubts about whether we need this separate interface. It's already included in `UFlowNodeBase` so all node types are using it. 

I might not see a reason for extracting this code to the interface. Is it perhaps useful in internal projects?
If there's no important reason for that, it would be worth moving this code back to the node class, as this would simplify understanding code for smaller teams and ease future maintenance.

Source/Flow/Public/Interfaces/FlowNativeExecutableInterface.h
MothDoctor: I have doubts about whether we need this separate interface. It's already included in `UFlowNode` and `UFlowNodeAddOn`, so perhaps it might be moved back to `UFlowNodeBase`?

I might not see a reason for extracting this code to the interface. Is it useful in internal projects?
If there's no important reason for that, it would be worth moving this code back to the node class, as this would simplify understanding code for smaller teams and ease future maintenance.

Source/Flow/Public/Nodes/FlowNode.h
MothDoctor: I see that is already fixed :)

Source/FlowEditor/Private/Graph/FlowGraphSchema_Actions.cpp
MothDoctor: Yes, but it is quite confusing to understand the editor code, as naming differs from the `AddOn` term. We also have SubGraphs in this plugin.

I would love to rename all "SubNode" occurrences to "AddOn" or "NodeAddOn". I'm only wondering if there would be some naming conflicts with other symbols. I don't think so, but might not see it whole picture yet ;)

LindyHopperGT: There's a number of instances in the UE code where there are multiple different types of SubNodes (in Behavior trees, for instance, there are Decorators and Services).  

I primarily am concerned with keeping the Editor-side code clear, so that we don't have "EdNodeNodes" and so on.  That's why i rebranded with AddOns for the runtime.  

But there is some value in keeping the Editor-side generalized to the idea that there may be other types of subnodes in the future, beyond AddOns, that would have a different runtime name.

Am not strongly opinionated on this though.  Something like "NodeAddOn" is fine editor-side, so long as the code reads clearly.

Let's let the code settle a bit before making those changes though.  Let me at least merge back the latest from Flow plugin before any renaming, so I don't have a crazy merge.

Source/Flow/Public/Interfaces/FlowCoreExecutableInterface.h
LindyHopperGT: We need "IFlowCoreExecutableInterface" and "IFlowExternalExecutableInterface" for the differences between internal and external flow executable cases.

As for the "IFlowNativeExecutableInterface", I scanned the code and couldn't find any compelling reason why it must be an interface as opposed to simply virtual functions on UFlowNodeBase.

I don't think we need to keep IFlowNativeExecutableInterface.

I can undo this, but I'd like to merge from Flow-mainline first to avoid a nasty merge.  And we'd have to change our code to suit, since we would need to replace references of IFlowNativeExecutableInterface with UFlowNodeBase references.

PR #206, opened 2024-05-18
Add a search functionality that does not rely on engine modifications.

dzxmxd:
### Description

I have implemented a new search functionality for the FlowGraph plugin that does not rely on engine modifications. This enhancement preserves the original plugin's features while improving compatibility and ease of integration. Key features include:
﻿
### Changes Made
1. **Independent Search Functionality**: Operates without engine modifications, ensuring compatibility with standard Unreal Engine setups.
2. **Node Description Search**: Added support to search within node descriptions, this is equivalent to searching for configuration content, which I think is very meaningful.
3. **SubGraph Search**: Supports searching within SubGraphs, with user control via a checkbox.

### Motivation
The previous search implementation required engine modifications, posing challenges for users preferring a standard engine setup. This update enhances flexibility and ease of use by decoupling the search functionality from engine changes.
﻿

Please review the changes and let me know if any further modifications are needed. The FlowGraph plugin has been incredibly helpful for my project. Thank you for your contribution to this fantastic plugin.

image

### DIFF
--- a/Source/FlowEditor/Private/Asset/FlowAssetEditor.cpp
+++ b/Source/FlowEditor/Private/Asset/FlowAssetEditor.cpp
@@ -31,6 +31,8 @@
 
 #if ENABLE_SEARCH_IN_ASSET_EDITOR
 #include "Source/Private/Widgets/SSearchBrowser.h"
+#else
+#include "Find/FindInFlow.h"
 #endif
 
 #define LOCTEXT_NAMESPACE "FlowAssetEditor"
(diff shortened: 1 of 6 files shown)


soraphis:
just took a look at the code, and I feel like this would be even better than the "Asset Search" plugin... 
Since the asset search plugin does not jump to the specific node (at least not for me).

So I'm wondering if it shouldn't be a straight up replacement, instead of #if/else 
(edit: ookay, saw the open PRs in the UE project from MothDoctor, so ... maybe a replacement until those are through?!)

MothDoctor:
It is a wonderful idea since Epic hasn't merged my pull requests yet.

The search mechanism used here is a custom logic, not using Asset Search db? If the end result isn't different from what users get while doing global Asset Search, I would gladly pull it in :)

dzxmxd:
Thank you for your feedback, MothDoctor!
﻿
The search functionality leverages a custom widget mechanism, with the logic inspired by FindInBT.h and FindInBT.cpp. Thus, its behavior is similar to the search results seen in Behavior Trees.
After seeing your reply, I took the time to compile the source engine. The functionality is mainly consistent. and in custom logic supports searching within SubGraphs and jumping directly to SubGraph nodes.
The search also indexes information from GetNodeDescription, which is frequently searched in my project. This feature is valuable as it allows searching for configuration content rather than just node names.
I’ve attached a GIF below demonstrating the actual effect. I hope this provides a clear picture of the effect.

soraphis:
what maybe might be nice: providing a function on the FlowNode base class "GetSearchKeywords" that returns an array of searchable keywords, for things that don't get added to the title or description!?

dzxmxd:
image

The current search contents are: NodeTitle, Description, NodeType, Node Comment.
From my experience, I feel that Description is similar to adding a new SearchKeywords. Of course, if it is for greater clarity, I am also happy to add this item

soraphis:
I'm currently using this PR in our project and so far it has been a nice upgrade over the unreal asset search plugin!

> From my experience, I feel that Description is similar to adding a new SearchKeywords. Of course, if it is for greater clarity, I am also happy to add this item

hm. some of our nodes have quite a lot of configurable data, that is not always shown in the description in text form since it would overload the description, or is written in natural language in the description, but that might not be how I'd search for it. 

But yeah, right now its working quite nicely so I would not complain if it would be added as is right now :)

LindyHopperGT:
We have some interest in this PR at Riot.  

Any idea when it can get merged to main?

MothDoctor:
Finally accepted! :) 

Also, fixed compilation issues against the Flow AddOn refactor, and corrected copyrights.

PR #207, opened 2024-06-05
Add a bunch of const keywords to allow usage of LogError/LogWarning/LogNote in const functions without const_cast

MaksymKapelianovych:


MothDoctor:

MaksymKapelianovych:
Updated!

Source/Flow/Private/Nodes/FlowNodeBase.cpp
```diff
@@ -612,12 +612,7 @@ void UFlowNodeBase::LogError(FString Message, const EFlowOnScreenMessageType OnS
 #endif
 }
 
-void UFlowNodeBase::LogErrorConst(FString Message, const EFlowOnScreenMessageType OnScreenMessageType) const
```
MothDoctor, line 635: I would keep separate const and const-version as it sometimes nodes need to modify some internal properties

Source/Flow/Private/Nodes/FlowNodeBase.cpp
```diff
@@ -633,12 +628,7 @@ void UFlowNodeBase::LogWarning(FString Message)
 #endif
 }
 
-void UFlowNodeBase::LogWarningConst(FString Message) const
```
MothDoctor, line 656: I would keep separate const and const-version as it sometimes nodes need to modify some internal properties

Source/Flow/Private/Nodes/FlowNodeBase.cpp
```diff
@@ -633,12 +628,7 @@ void UFlowNodeBase::LogWarning(FString Message)
...
-{
-	const_cast<UFlowNodeBase*>(this)->LogWarning(Message);
-}
-
-void UFlowNodeBase::LogNote(FString Message)
+void UFlowNodeBase::LogNote(FString Message) const
```
MothDoctor, line 651: I would keep separate const and const-version as it sometimes nodes need to modify some internal properties

MothDoctor:
One more merge fix will be needed here. I will pull as the next PR after merge fix and applying my comments :)

Source/Flow/Private/Nodes/FlowNodeBase.cpp
MaksymKapelianovych: I am afraid I do not understand your point here... LogError and LogErrorConst can be used in both const and non-const functions (if this PR will be merged), they itself do not modify any properties. Could you please give an example where I can use LogErrorConst, but not LogError?

MothDoctor: OK :)

PR #210, opened 2024-06-16
Improvements for FlowGraphDiffing

LindyHopperGT:
- fix FlowGraph details diffing by by assigning Splitter to the FDiffControl.Widget in SFlowDiff::GenerateDetailsPanel()
- Add nesting to tree entry display to more easily read the diffs.
- Add individual property diffing with highlights.
- Nest Add-On nodes inside their parent flow nodes
- added FlowObjectDiff.h/cpp
- Added FFlowObjectDiff to track all the data needed to display an individual tree entry's diff.


Source/Flow/Private/Nodes/FlowNode.cpp
```diff
@@ -112,7 +112,7 @@ void UFlowNode::AddOutputPins(TArray<FFlowPin> Pins)
 
 bool UFlowNode::RebuildPinArray(const TArray<FName>& NewPinNames, TArray<FFlowPin>& InOutPins, const FFlowPin& DefaultPin)
 {
-	bool bIsChanged = false;
```
LindyHopperGT: Oh.  I didn't mean to change this from mainline.  I thought mainline removed the false.  ('Resharper' will sometimes remove default values like this if all 'if' branches set the value)

I'll put it back in.

MothDoctor: Yes, I removed the false as static analysis pointed that out.

Source/Flow/Private/AddOns/FlowNodeAddOn.cpp
```diff
@@ -68,7 +68,14 @@ bool UFlowNodeAddOn::IsSupportedInputPinName(const FName& PinName) const
 		return true;
 	}
 
-	return FindFlowPinByName(PinName, InputPins) ? true : false;
+	if (const FFlowPin* FoundFlowPin = FindFlowPinByName(PinName, InputPins))
+	{
+		return true;
```
LindyHopperGT, line 73: Broke this back out into if/else so we can set breakpoints.

Source/Flow/Private/Nodes/World/FlowNode_ExecuteComponent.cpp
```diff
@@ -217,53 +215,53 @@ bool UFlowNode_ExecuteComponent::TryInjectComponent()
 	static_assert(static_cast<int32>(EExecuteComponentSource::Max) == 4, TEXT("Update this code if the enum changes"));
 	switch (ComponentSource)
 	{
-		case EExecuteComponentSource::InjectFromTemplate:
+	case EExecuteComponentSource::InjectFromTemplate:
```
LindyHopperGT, line 218: Fixing indenting regression

Source/Flow/Private/Nodes/World/FlowNode_ExecuteComponent.cpp
```diff
@@ -316,24 +314,25 @@ const UActorComponent* UFlowNode_ExecuteComponent::TryGetExpectedComponent() con
...
-		case EExecuteComponentSource::InjectFromTemplate:
-			{
-				return ComponentTemplate;
-			}
-		case EExecuteComponentSource::InjectFromClass:
-			{
-				return IsValid(ComponentClass) ? ComponentClass->GetDefaultObject<UActorComponent>() : nullptr;
-			}
-		default:
+	case EExecuteComponentSource::Undetermined:
```
LindyHopperGT, line 317: Reintroduced reverted whitespace

Source/Flow/Private/Nodes/World/FlowNode_ExecuteComponent.cpp
```diff
@@ -479,46 +478,47 @@ FText UFlowNode_ExecuteComponent::GetNodeTitle() const
 		static_assert(static_cast<int32>(EExecuteComponentSource::Max) == 4, TEXT("Update this code if the enum changes"));
 		switch (ComponentSource)
 		{
-			case EExecuteComponentSource::Undetermined:
-				break;
+		case EExecuteComponentSource::Undetermined:
```
LindyHopperGT, line 481: fixed indention regression

Source/Flow/Private/Nodes/World/FlowNode_ExecuteComponent.cpp
```diff
@@ -538,40 +538,43 @@ void UFlowNode_ExecuteComponent::UpdateNodeConfigText_Implementation()
 		static_assert(static_cast<int32>(EExecuteComponentSource::Max) == 4, TEXT("Update this code if the enum changes"));
 		switch (ComponentSource)
 		{
-			case EExecuteComponentSource::Undetermined:
-				break;
-			case EExecuteComponentSource::BindToExisting:
+		case EExecuteComponentSource::Undetermined:
```
LindyHopperGT, line 541: Fixed intention regression and whitespace

Source/Flow/Public/Nodes/FlowNodeBase.h
```diff
@@ -63,6 +63,27 @@ class FLOW_API UFlowNodeBase
 	virtual void Cleanup() override;
 	// --
 
+	// Finish execution of node, it will call Cleanup
```
LindyHopperGT, line 66: IFlowNativeExectableInterface functions folded into UFlowNodeBase.

Source/Flow/Public/Nodes/FlowNodeBase.h
```diff
@@ -28,7 +28,7 @@ typedef TFunction<void(UFlowNodeAddOn&)> FFlowNodeAddOnFunction;
 /**
  * The base class for UFlowNode and UFlowNodeAddOn, with their shared functionality
  */
-UCLASS(Abstract, HideCategories = Object)
+UCLASS(Abstract, BlueprintType, HideCategories = Object)
```
LindyHopperGT, line 31: Must be BlueprintType if UFlowNodeBase will serve the IFlowNativeExecutableInterface role

Source/FlowEditor/Private/DetailCustomizations/Pre_5_3_GetActorClassDefaultComponents.h
```diff
@@ -0,0 +1,82 @@
...
+
+#if USE_PRE_5_3_GET_ACTOR_CLASS_DEFAULT_COMPONENTS
+
+#include "Components/ActorComponent.h"
+#include "Engine/BlueprintGeneratedClass.h"
+#include "Engine/SCS_Node.h"
+#include "Engine/SimpleConstructionScript.h"
+#include "GameFramework/Actor.h"
+
+FORCEINLINE_DEBUGGABLE void Pre_5_3_ForEachComponentOfActorClassDefault(const TSubclassOf<AActor>& ActorClass, const TSubclassOf<UActorComponent>& InComponentClass, TFunctionRef<bool(const UActorComponent*)> InFunc)
```
LindyHopperGT, line 16: Strictly a copy/paste of AActor functions from 5.3 to add pre-5.3 compatibility for this function

Source/Flow/Private/AddOns/FlowNodeAddOn.cpp
MaksymKapelianovych: Why not do it like this?
```
bool bFoundPin = FindFlowPinByName(PinName, InputPins) != nullptr;
return bFoundPin;
```
Or this? 
```
bool bFoundPin = !!FindFlowPinByName(PinName, InputPins);
return bFoundPin;
```

LindyHopperGT: Those are more efficient than the ternary approach (that is, no branch).

But none of them allow you to set a breakpoint on the return statement that is unique for the 'return true' separate from the 'return false' case.

It's something I like to have when I'm debugging (being able to set a breakpoint on either the return true or return false case, in isolation).  But if we don't expect to need to do that, then your first option is the preferred way in my estimation.

Personally, I'd like either my version, with if/else, or your first version (bool bFoundPin = FindFlowPinByName(PinName, InputPins) != nullptr;)

but not the ternary or the !! version.

MaksymKapelianovych: You can always set conditional breakpoint to achieve the same behaviour
image

LindyHopperGT: It's faster to set a regular breakpoint.

I'm not particularly concerned either way.

---
The standard I have set for my engineers is to prefer if/else over ternary for anything that we would ever want to set a breakpoint on either branch result (that is, only use ternary for trivial cases like null-protection, or setting a constant variable, etc.)

Otherwise, we suggest avoiding a branches when possible.

I saw this ternary as something that should be shifted in one way or the other:  either it should be an if/else, or it should be branchless bool return.

That's my reasoning, and why I reverted the code back to if/else.  I prefer the breakpoint option that if/else gives over ternary.

Source/Flow/Private/Nodes/World/FlowNode_ExecuteComponent.cpp
MothDoctor: Heh, this was done to conform to plugin style which is - I need to recheck that - the same as the engine standard. I won't be touching this now so you won't have annoying merge conflicts while working on this feature..

In the future, I would re-check the plugin style and adjust everywhere to the same style. I have no personal preference here, other than consinstency.

PR #212, opened 2024-06-28
fix[PlayLevelSequence Node]: Prevent multiple output pins with same name

soraphis:
e.g. adding multiple notifies to a sequence named "Pause" would previously add multiple output pins.
    
We have a level sequence that pauses at multiple points during the sequence, it has a "Pause" notify that is looped back into the pause input of the flow node. 

Currently all these Pause events are creating multiple Context Output Pins with the same name.


MothDoctor:
Awesome, thank you for the fix :)

PR #213, opened 2024-07-04
Restore SFlowGraphNode pin alignment to pre-FlowAddon settings

Unbansheee:
PR #202 added an override for SFlowGraphNode's AddPin function, causing pin alignment to be incorrect. This PR simply replaces the alignment and padding settings in AddPin with the original alignment settings in SGraphNode (default for pre-FlowAddons)

Alignment at present:
Screenshot 2024-07-04 175249

Alignment and padding restored:
Screenshot 2024-07-04 175121


MothDoctor:
Thank you for the fix! Merged :)

PR #214, opened 2024-07-05
feat: Added BlueprintCallable function to trigger a custom input on a FlowComponent.

soraphis:
this allows more control from the owner of a FlowComponent, incases where sending a notify from actor would be more cumbersome.


MothDoctor:
Thanks for the submission. I accepted it and updated the naming of previously existing methods related to Custom Outputs, so it's now consistent.

PR #223, opened 2024-08-17
Fix compile failure due to default C++ standard below 20

Lim-Young:
For some projects upgrading from older versions to UE5.2 or later, the default C++ standard used by the engine may not be C++20. This issue arises because the file ReviewComments.h contains syntax that is only supported in C++20:

```cpp 
bool bIsClosed : 1 = false;
```

This bitfield initialization will cause the module compilation to fail. Adding logic to ensure C++ standards in the module's Build.cs can solve this problem.


MothDoctor:
Thanks for the fix, accepted! :)

PR #224, opened 2024-08-19
Fix compilation for iOS

slobodin:
Fixed few issues that were preventing compilation of the plugin on iOS (clang)


MothDoctor:
Thanks for the fix, accepted! :)

PR #227, opened 2024-08-26
Add missing serialization calls for saved data structs

dyanikoglu:
Not sure why those calls were not included in serialize functions.


rweber89:
@dyanikoglu this change doesnt include a version for the manual serialization, so not all changes in newer version will be backwards compatible.

MothDoctor:
Indeed, this might break loading already created saves in released games.
Is there a specific reason why you proposed these changes? I understand that current code feels incomplete, but works fine with properties marked as `SaveGame`?

dyanikoglu:
In my custom save system, I'm serializing those structs manually and writing into saved data. If it's likely break compatibility with existing projects, the PR can be closed.

PR #228, opened 2024-08-26
Make it possible to pass InstanceName to CreateRootFlow

dyanikoglu:
This is a useful parameter for custom save/load systems integrated with Flow. I needed this change to provide old instance name from save data while restoring root flow. Otherwise, it gets instantiated as FooFlowAsset_1, FooFlowAsset_2, FooFlowAsset_3 ... etc on each game load.


MothDoctor:
Merged it up! Thank you! :)

PR #229, opened 2024-08-28
Add virtual to StartRootFlow & FinishRootFlow

dyanikoglu:
Makes it easier to extend UFlowComponent for project specific needs.


MothDoctor:
Thank you for your contribution! Accepted :)

PR #234, opened 2024-10-04
Extension point to allow FlowGraphNode to do additional indexing.

SPontadit:
Add a virtual function AdditionalNodeIndexing in FlowGraphNode to index property that are not handled by default


MothDoctor:
Thank you for your contribution! Accepted :)

PR #235, opened 2024-10-04
Call NotifyGraphChanged when validating an asset to refresh nodes without making the asset dirty

SPontadit:
When ValidateAsset is called, the nodes are not reconstructed, so the ErrorReporting widget is not updated. We have to call Refresh to update the node but, it makes the asset dirty.

By calling NotifyGraphChanged  at the end of ValidateAsset_Internal, nodes are reconstructed without marking the asset dirty.


MothDoctor:
Thank you for a change. I tested it internally and found no regressions, so it had been accepted! :)

PR #241, opened 2024-11-21
Initialize Data Pin Struct Values in order to remove errors

Benxidosz:
There were errors like this: "StructProperty FFlowDataPinResult_Rotator::Value is not initialized properly even though its struct probably has a custom default constructor. Module:Flow File:Public/Types/FlowDataPinResults.h". I initialized the problematic properties, so all the errors disappeared.


MothDoctor:
Thanks for the fix! Merged :)

PR #244, opened 2024-11-26
Extending Allow/Deny to include AddOns

LindyHopperGT:
- also adding capability to opt back-in to being allowed as a subclass of a denied node


MothDoctor:
Oh, nice. Change accepted! Thank you :)

PR #246, opened 2024-11-28
Replace SFlowGraphNode style with default K2Node style.

Benxidosz:
This was needed because style plugins like "Darker Nodes" couldn't affect these nodes.

(cherry picked from commit 0b502276de03e8dac2b73c5e86dabce76f37a978)


MothDoctor:
Well, if I read the change correctly, it would remove custom Flow styling, right?
In this case, I'm afraid I can't accept this change.

Benxidosz:
Yes, but because of the custom style, plugins like Darker nodes, can't affect them. Although some part of the flow nodes using the default engine style like the shadow brush (see SFlowGraphNode.cpp::98), this is the selection highlight. Because of that the selection highlight will be affected by the Darker nodes, which can led to a broken layout. I had a problem, where the selection was behind the node.

If you want to keep the flow custom style, I recommend to make a custom style for this selection as well. In that case nothing will react to these style plugins, so they won't break the layout.

PR #251, opened 2024-12-18
Fix compile errors against more strict -Wmissing-declarations

Ryan-DowlingSoka:


MothDoctor:
Thank you for that! I will look into adding that compilation flag to my build system tests :)

PR #252, opened 2024-12-19
Enable plugin requirement EngineAssetDefinitions it relies on

Ryan-DowlingSoka:
```Warning: Plugin 'Flow' does not list plugin 'EngineAssetDefinitions' as a dependency, but module 'FlowEditor' depends on module 'EngineAssetDefinitions'.```

### DIFF
--- a/Flow.uplugin
+++ b/Flow.uplugin
@@ -35,6 +35,10 @@
 			"Enabled": true
 		},
 		{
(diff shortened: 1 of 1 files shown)


MothDoctor:
Accepted, thanks! I already forgot twice about fixing that ;)

PR #255, opened 2024-12-28
Graph refresh system cleanup

HomerJohnston:
The goal of this PR is to make the graph editor stop refreshing the entire graph during as many events as possible. It also fixes a few minor bugs related to orphaned pins, running undo/redo commands on edge cases, and unnecessary graph dirtying on edge cases such as when nodes contain orphaned or invalid pins. 

Undo/redo events unfortunately still rebuild the whole graph (Blueprint graphs seem to do this also; I may check later if there is a way to work around this safely or not).

This is a fairly large PR; I have tested this in a small zoo test graph with lots of different pin setups as much as I can, but some additional testing in a "real world" graph would be wise.

**Breaking Changes:**

1. `public void UFlowGraphNode::RefreshContextPins(bool)` changed to `protected void UFlowGraphNode::RefreshContextPins()`
   - The existing system has some logic which could result in attempts to recursively call `ReconstructNode -> RefreshContextPins -> ReconstructNode -> RefreshContextPins`, currently there are bools to guard against this, but both methods are exposed and this felt unclean to me. My suggestion is to protect `RefreshContextPins` and leave `ReconstructNode` for public usage. Both functions have been reworked.

**Other Major Changes:**

2. `UFlowGraphNode::PostLoad` - now rebuilds the unserialized `InputPins` and `OutputPins` arrays using the serialized `Pins` array.
3. `UFlowAsset::HarvestNodeConnections` - Made this function able to harvest a single node, or harvest the whole graph (pass in nullptr to harvest the whole graph, or pass in a single node to harvest just that node). I then use `UFlowGraphNode::NodeConnectionListChanged` to update node connections for only a single node when its connections are changed.
4. Changed the right-click context menu command, "Refresh Context Pins", to "Reconstruct Node".
5. Modified the following functions to only refresh individual nodes instead of the whole graph by changing `NotifyGraphChanged()` calls to `NotifyNodeChanged(Node)`:
   - `FFlowGraphSchemaAction_NewNode::CreateNode`
   - `UFlowGraphSchema::TryCreateConnection`
   - `UFlowGraphSchema::BreakNodeLinks`
   - `UFlowGraphNode::RemoveOrphanedPin`
   - `UFlowGraphNode::AddInstancePin`
   - `UFlowGraphNode::RemoveInstancePin`
6. `FFlowGraphSchemaAction_NewNode::CreateNode` - now runs `ReconstructNode()` instead of only spawning default pins.
7. `UFlowGraphNode::OnGraphRefresh` - now runs `ReconstructNode()` instead of `RefreshContextPins()`, on all nodes.
8. `UFlowGraphNode::RefreshContextPins` - no longer runs at all if an editor transaction is in progress (running undo/redo).
9. `UFlowGraphNode::HavePinsChanged` - now disregards any orphaned pin connections.
10. Added a new delegate `UFlowGraphNode::OnReconstructNodeCompleted` to trigger rebuild of `SFlowGraphNode` widgets whenever ReconstructNode is called.
11. `SFlowGraphNode` - added destructor to unbind from UFlowGraphNode delegates.
12. `UFlowGraphSchema` - added `virtual bool ShouldAlwaysPurgeOnModification() const override { return false; }` to reduce unnecessary graph refreshes.
13. `UFlowNodeAddOn` - GetContextInput/OutputPins functions changed to ignore invalid pins and write a log warning for invalid pins.

### DIFF
--- a/Source/Flow/Private/AddOns/FlowNodeAddOn.cpp
+++ b/Source/Flow/Private/AddOns/FlowNodeAddOn.cpp
@@ -1,6 +1,8 @@
 // Copyright https://github.com/MothCocoon/FlowGraph/graphs/contributors
 
 #include "AddOns/FlowNodeAddOn.h"
+
+#include "FlowLogChannels.h"
 #include "Nodes/FlowNode.h"
 
 #include "Misc/RuntimeErrors.h"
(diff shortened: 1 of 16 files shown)


MothDoctor:
Huge thanks for that! I did try in a pet project for a few days, but that's not enough to fully test such a refactor. So I'm accepting and letting know the community. Even if some issues are detected, your change already fixes some previously reported issues.

Thank you for the contribution <3

PR #256, opened 2025-01-06
Post-merge updates from Riot version

LindyHopperGT:
DeinitializeInstance() - safety for calling when already deinitialized

Recategorize flow nodes:
- Developer - Log
- Flow|Graph - Custom Input, Custom Output, Define Properties, Finish, Subgraph
- Flow|Owner - Execute Component
- Flow|Routing - Branch, Counter, Do Once, Multigate, Reroute, Sequence, Timer*
- Flow|Routing|Logic - And, Or
- Flow|Save - Checkpoint
- World - Play Level Sequence
- World|Actor - OnActorRegistered, OnActorUnregistered
- World|EQS - RunEQSQuery (AI Flow)
* Not sure about the categorization for Timer, but I kept it in 'routing' because it was in Route category before

* This recategorization was a part of our own nodes being recategorized to improve designer workflows.  It's really more of a "what about this?" pass on the categories than a necessity.  Definitely open to reverting or revising this part of the PR.

Restored signature for creating a flow graph with a custom schema

Some clerical edits


MothDoctor:
> Definitely open to reverting or revising this part of the PR.

Actually, it was easy to add a new project setting allowing users to override node categories as they like. It requires adding only a small method to read this setting.
https://github.com/MothCocoon/FlowGraph/commit/53170b4a56392d21f032fa04b21ced723f877120

Here I put your category layout. And the visual effect in the Flow Solo sample.

image
image

I'm dropping .ini entry, so you won't have to re-do it ;)
`(("/Script/CoreUObject.Class'/Script/Flow.FlowNode_LogicalAND'", "Flow|Routing|Logic"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_LogicalOR'", "Flow|Routing|Logic"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_Branch'", "Flow|Routing"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_Counter'", "Flow|Routing"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_CustomInput'", "Flow|Graph"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_CustomOutput'", "Flow|Graph"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_ExecutionMultiGate'", "Flow|Routing"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_ExecutionSequence'", "Flow|Routing"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_Finish'", "Flow|Graph"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_Reroute'", "Flow|Routing"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_Start'", "Flow|Graph"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_SubGraph'", "Flow|Graph"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_Timer'", "Flow|Routing"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_Checkpoint'", "Flow|Save"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_Log'", "Developer"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_ExecuteComponent'", "Flow|Owner"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_OnActorRegistered'", "World|Actor"),("/Script/CoreUObject.Class'/Script/Flow.FlowNode_OnActorUnregistered'", "World|Actor"))`

MothDoctor:
Now, I admit it this nested layout looks tempting. It might make sense to change C++ defaults, even though users can now override them.

However, I wouldn't pull current changes to node defaults.
* I see a few opportunities to improve categorization before changing anything. Example: Notify nodes should be also listed under the same categories as Actor (Un)Registered Nodes. Perhaps we could have only an `Actor` category instead of `World/Actor`? Probably we might still skip the `Flow` category part, but... it seems that nested categories are useful in your project. So... not sure if we could reach a consensus satisfying most of the projects "by default". After all, the current simple form lasted for a few years.
* As a general rule, such a change should be introduced as a separate pull request. (Yeah, I know, I should write down some PR rules, guilty as charged).

So... I would propose to remove node Category changes from this PR, and I'd accept the rest of the integration. We might continue to discuss the topic of categories in a separate PR. Or we can drop it, if your team needs will be satisfied by the ability to override categories in Project Settings. Both options are fine to me :)

Anyway, the community already got a new option thanks to your proposal! Thank you! :)

MothDoctor:
I merged that in. I did partially accept a proposed update to the category layout. I updated the defaults, which are more flat than proposed, as this would be enough for projects starting with a limited number of project-specific nodes.

**BREAKING CHANGE**. Flow Node source files have been moved to matching folder names. You might need to update your includes if you have C++ classes referencing built-it Flow Node classes in your code.

image

PR #258, opened 2025-01-14
Small addition for custom events to help find them in a large graphs. 

MaksymKapelianovych:
Added "Select" button (see image) which will select and focus Custom Event node with the same name. If such node does not exist in the graph, button will be inactive.
image

### DIFF
--- a/Source/Flow/Private/FlowAsset.cpp
+++ b/Source/Flow/Private/FlowAsset.cpp
@@ -966,11 +966,14 @@ void UFlowAsset::RemoveCustomOutput(const FName& EventName)
 
 UFlowNode_CustomInput* UFlowAsset::TryFindCustomInputNodeByEventName(const FName& EventName) const
 {
-	for (UFlowNode_CustomInput* InputNode : CustomInputNodes)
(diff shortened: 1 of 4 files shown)


MaksymKapelianovych:
P.S.
Honestly, I would like to completely rework implementation of CustomEvent nodes and make working with them more more like in BP. Simple right-click and "Add Custom Input/Output" is more convenient than editing two separate arrays of FNames, than create Event nodes and linking them together.
It should look something like this (image) instead of current arrays, but keep track of both inputs and outputs. 
image 

This approach has some other minor issues: 
 - arrays can have entries that do not exist in a graph and there is no way to tell user about this (until this PR)
 - renaming array element do not affect existing node in a graph. To fix this we should introduce some kind of linking between array entries and nodes, and, probably, wrap it in a custom struct. This will lead to a need for a way to migrate existing arrays to new implementation in order to not break existing assets.
 - maybe other issues I am not aware of

What do you think about the issues and the need to fix them?

P.P.S.
I haven't come up with a solution of how to keep backwards compatibility between two different implementations, so new PR will not come soon (

PR #259, opened 2025-01-16
Updates for AI Flow GetBlackboardValues

LindyHopperGT:
Added additional constructors for FFlowDataPinResult_Enum(), that are used in an AI Flow node: "GetBlackboardValues"

Moved the RefreshContextPins() call to before ShouldReconstructNode(), so that it has an opportunity to rebuild the context pins source values prior to testing if the node needs to be rebuilt (which tests the context pin source values vs the pin widgets for differences).


Source/Flow/Public/Nodes/FlowPin.h
```diff
@@ -20,14 +20,14 @@ struct FLOW_API FFlowPin
 	GENERATED_BODY()
 
 	// A logical name, used during execution of pin
-	UPROPERTY(EditDefaultsOnly, Category = "FlowPin")
+	UPROPERTY(EditDefaultsOnly, BlueprintReadWrite, Category = "FlowPin")
```
LindyHopperGT: It turns out we needed these exposed to blueprint so we can make FlowPin structs and set their values in GetContextPins calls in Blueprint.  

These are only called in editor, but they still are called from blueprint.

Source/Flow/Public/Nodes/FlowPin.h
```diff
@@ -20,14 +20,14 @@ struct FLOW_API FFlowPin
...
 	FName PinName;
 
 	// An optional Display Name, you can use it to override PinName without the need to update graph connections
-	UPROPERTY(EditDefaultsOnly, Category = "FlowPin")
+	UPROPERTY(EditDefaultsOnly, BlueprintReadWrite, Category = "FlowPin")
```
LindyHopperGT: We could leave this one out if it's a problem.  We don't technically need to write to PinFriendlyName from blueprint.

MaksymKapelianovych: Maybe it would be better to create native make/break functions and leave these properties without BlueprintReadWrite?

MaksymKapelianovych: And it would also allow to wrap PinFriendlyName and PinToolTip in WITH_EDITORONLY_DATA

Source/FlowEditor/Private/DetailCustomizations/FlowDataPinProperty_ClassCustomization.cpp
```diff
@@ -164,7 +164,8 @@ const UClass* FFlowDataPinProperty_ClassCustomizationBase::OnGetClass() const
 
 UClass* FFlowDataPinProperty_ClassCustomizationBase::BuildMetaClass() const
 {
-	return CachedMetaClassPtr.Get();
+	UClass* MetaClass = CachedMetaClassPtr.Get();
```
LindyHopperGT, line 167: Crash fix for UE 5.5, as discussed on Discord General Chat

LindyHopperGT:
I updated PR 259 with make/break static functions for blueprint and hid the properties from blueprint read/write. I didn’t make them editor only data because I dont have time to follow through with the extra fixes that change would entail. But it’s something we could consider in future to slim up these structs in deployed packages. Cheers,Greg TaylorOn Jan 16, 2025, at 2:25 PM, MaksymKapelianovych ***@***.***> wrote:﻿
@MaksymKapelianovych commented on this pull request.



In Source/Flow/Public/Nodes/FlowPin.h:
>  	FName PinName;
 
 	// An optional Display Name, you can use it to override PinName without the need to update graph connections
-	UPROPERTY(EditDefaultsOnly, Category = "FlowPin")
+	UPROPERTY(EditDefaultsOnly, BlueprintReadWrite, Category = "FlowPin")

And it would also allow to wrap PinFriendlyName and PinToolTip in WITH_EDITORONLY_DATA

MaksymKapelianovych:
I would also suggested to add 'meta = (HasNativeMake = "/Script/Flow.FlowDataPinBlueprintLibrary.MakeStruct", HasNativeBreak = "/Script/Flow.FlowDataPinBlueprintLibrary.BreakStruct")' to FFlowPin, because with current setup you will still have a standard generated make/break alongside with custom functions. 
Meta will force editor to use custom functions instead.

MothDoctor:
Accepted changes :) 
@MaksymKapelianovych also included your suggestion :)

PR #262, opened 2025-02-06
SFlowGraphPinExec with Custom "PinColorModifier":

Benxidosz:
+ Added PinColorModifier FLinearColor Property into FFlowPin -> With this we can choose the Color Modifier of a Pin in Runtime module
+ FlowNode stores the "PinColorModifier" value in a map for each PinName -> PinFactory can access this information
+ New "PinModifierColor" Slate Argument in SFlowGraphPinExec
+ Pin Factory set the "PinModifierColor" Argument when Construct the Pin

(cherry picked from commit 43af73f0da9436aaccd566bed82fa526367bfc0e)


MothDoctor:
Hey, here I finally decided not to integrate this change. I can't see a benefit here. It might actually be confusing to the general community since we now have Data Pins,  AddOns. This one should be kept as a modification done in specific projects.

PR #263, opened 2025-02-12
RemoveSubFlow - Set NodeOwningThisAssetInstance  after FinishFlow

pixelchamp:
Make sure to set the NodeOwningThisAssetInstance after the FinishFlow call, as it may be needed in the FinishFlow method.
We need to be able to access the owning node in FinishFlow call.


MothDoctor:
Oh, nice found. Merged! :)

PR #264, opened 2025-02-15
UFlowGraphNode::ReconstructNode and associated functions rewrite to clean up graph behavior

HomerJohnston:
- Removed some duplicate calls to reconstruct nodes; `ReconstructNode()` would frequently run 2 or 3 times before, usually only runs once now
- Added additional flags to skip `ReconstructNode()`: no longer reconstructs before saving the graph, no longer reconstructs a node before deleting the node
- `ReconstructNode()` completely rewritten to clean up logic and resolve bugs with 'false' graph asset dirtying or missed pin change updates
- `UFlowGraphNode::RefreshContextPins()` function completely replaced by two new substantial functions: 
  - `TryUpdateAutoDataPins()`
  - `TryUpdateNodePins()`

### DIFF
--- a/Source/FlowEditor/Private/Graph/FlowGraph.cpp
+++ b/Source/FlowEditor/Private/Graph/FlowGraph.cpp
@@ -93,15 +93,8 @@ void UFlowGraph::RefreshGraph()
 			}
 		}
 
+		// This function will (eventually) result in all graph nodes being reconstructed
 		UnlockUpdates();
-	}
-
-	// refresh nodes
-	TArray<UFlowGraphNode*> FlowGraphNodes;
-	GetNodesOfClass<UFlowGraphNode>(FlowGraphNodes);
-	for (UFlowGraphNode* GraphNode : FlowGraphNodes)
-	{
-		GraphNode->OnGraphRefresh();
 	}
 }
 
@@ -222,7 +215,11 @@ void UFlowGraph::OnLoaded()
 
 void UFlowGraph::OnSave()
 {
+	bIsSavingGraph = true;
+	
 	UpdateAsset();
+
+	bIsSavingGraph = false;
 }
 
 void UFlowGraph::Initialize()
--- a/Source/FlowEditor/Private/Graph/Nodes/FlowGraphNode.cpp
+++ b/Source/FlowEditor/Private/Graph/Nodes/FlowGraphNode.cpp
@@ -292,7 +292,7 @@ void UFlowGraphNode::ReconstructNode()
 	}
 	
 	bIsReconstructingNode = true;
-	FScopedTransaction Transaction(LOCTEXT("ReconstructNode", "Reconstruct Node"));
+	FScopedTransaction Transaction(LOCTEXT("ReconstructNode", "Reconstruct Node"), !GUndo);
 
 	bool bNodeDataPinsUpdated = TryUpdateAutoDataPins(); // This must be called first, it updates the underlying data for data pins of the flow node 
 	bool bNodeExecPinsUpdated = TryUpdateNodePins(); // Updates all pins of the flow node (native pins, meta auto pins, and context pins which include data pins for now)
@@ -322,10 +322,6 @@ void UFlowGraphNode::ReconstructNode()
 		}
 	
 		bNeedsFullReconstruction = false;
-	}
-	else
-	{
-		Transaction.Cancel();
 	}
 
 	// This ensures the graph editor 'Refresh' button still rebuilds all of the graph widgets even if the FlowGraphNode has nothing to update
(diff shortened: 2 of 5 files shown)


Source/FlowEditor/Private/Graph/FlowGraph.cpp
```diff
@@ -93,16 +93,9 @@ void UFlowGraph::RefreshGraph()
 			}
 		}
 
+		// This function will (eventually) result in all graph nodes being reconstructed
 		UnlockUpdates();
```
MaksymKapelianovych, line 97: I think it is better to remove UpdateAsset() call from UnlockUpdated(), because it's unclear that this function does this (unless you specifically look into the function body) and add it instead of deleted code below. 
And actually instead of having two functions just to change one bool value I would write `TGuardValue<uint32> LockUpdatesGuard(bLockUpdates, true);` at the beginning of the corresponding block (or even simpler `bLockUpdates = true/false`). Or provide a better name for UnlockUpdates(), for example, UnlockUpdatesAndUpdateAsset() (too mouthful and clearly states two separate responsibilities of this function).

Source/FlowEditor/Private/Graph/Nodes/FlowGraphNode.cpp
```diff
@@ -1784,36 +1664,213 @@ void UFlowGraphNode::ValidateGraphNode(FFlowMessageLog& MessageLog) const
...
 	{
 		return false;
 	}
 	
-	// If the graph is locked, we shouldn't reconstruct nodes 
-	// (all nodes will all be reconstructed when the graph is unlocked)
+	// Corrupt? This should never happen
+	if (!IsValid(NodeInstance))
+	{
+		UE_LOG(LogFlow, Error, TEXT("FlowGraphNode has no NodeInstance, graph may be corrupt!"))
```
MaksymKapelianovych: Maybe add more info about the node being corrupt and also add `ensureMsgf(Always)`?

Source/FlowEditor/Private/Graph/Nodes/FlowGraphNode.cpp
```diff
@@ -286,48 +286,49 @@ void UFlowGraphNode::InsertNewNode(UEdGraphPin* FromPin, UEdGraphPin* NewLinkPin
...
-	TArray<UEdGraphPin*> OldPins(Pins);
-	
-	// Harvest the auto-generated pins before refreshing context pins
-	if (UFlowNode* FlowNode = Cast<UFlowNode>(NodeInstance))
+	bool bGraphNodeRequiresReconstruction = bNeedsFullReconstruction || bNodeDataPinsUpdated || bNodeExecPinsUpdated || bAreGraphPinsMismatched;
+
+	if (bGraphNodeRequiresReconstruction)
 	{
-		if (UFlowAsset* FlowAsset = NodeInstance->GetFlowAsset())
+		Modify();
```
MaksymKapelianovych, line 305: Shouldn't this call be before TryUpdateAutoDataPins()?

Source/FlowEditor/Private/Graph/Nodes/FlowGraphNode.cpp
```diff
@@ -1784,36 +1664,213 @@ void UFlowGraphNode::ValidateGraphNode(FFlowMessageLog& MessageLog) const
...
+{
+	if (LeftPins.Num() != RightPins.Num())
 	{
 		return false;
 	}
+	
+	for (const FFlowPin& Left : LeftPins)
+	{
+		// For each required pin, make sure the existing pins array contains a pin that matches by name and type 
+		if (!RightPins.ContainsByPredicate( [&Left] (const FFlowPin& Right)
```
MaksymKapelianovych: I would move lambda out of if statement - easier to read.

Source/FlowEditor/Private/Graph/Nodes/FlowGraphNode.cpp
HomerJohnston: Not necessary because the flow graph node isn't modified until this point (and it doesn't really matter when exactly you call Modify, as long as it gets called sometime during the process), but!! You made me realize I should have called Modify() on the FlowNode, elsewhere in my new functions. Thanks :)

Source/FlowEditor/Private/Graph/FlowGraph.cpp
HomerJohnston: I very much agree in principle! However, the UnlockUpdates() code basically connects directly into the new "Addons" code, and I haven't studied the addons code at all yet (I don't use any addons). So I think this topic might make a good future PR for further cleanup, but I'm not sure if I want to risk expanding this PR into it. It would take me a while to make sure I'm comfortable with it (might not have more free time for a few weeks).

(PS - thanks for the review!)

Source/FlowEditor/Private/Graph/Nodes/FlowGraphNode.cpp
MaksymKapelianovych: I missed that pins are being changed in FlowNodeInstance and not in the UFlowGraphNode, but yeah, I meant that)

Issue #266, opened 2025-02-16
node pass through does not work under closed-loop conditions


nozomanai:
After testing, it was found that when a line forms a loop, passing through will trigger error.

Example node：

![Image](https://github.com/user-attachments/assets/67b93426-4834-49d2-8704-2f4eb05b5148)

I pass through IncRound node，get the error：
![Image](https://github.com/user-attachments/assets/9c6699db-ddcd-42f9-90c2-0e3fe636a10e)

no loop, work fine：
![Image](https://github.com/user-attachments/assets/19b7ecc7-c69a-403f-a841-1aa4138fd334)

MothDoctor:
Integrated Maksym's fix :)
https://github.com/MothCocoon/FlowGraph/pull/267

PR #267, opened 2025-02-16
Fix PassThrough nodes in loops

MaksymKapelianovych:
Fix for issue https://github.com/MothCocoon/FlowGraph/issues/266


MothDoctor:
Awesome, thanks for the fix! :)

PR #274, opened 2025-02-27
Properly handle disabling of ViewportStatsSubsystem is UFlowNodeBase/…

jnucc:
…UFlowComponent is being destroyed.

On our production, some LogError() was called in UFlowNode_Timer where the input 'In' was called more than once, triggering "Timer already active". Later on, as it was completed inside a flow, UFlowNode_Timer gets garbage collected. The delegate in UViewportStatsSubsystem* would still be called regularly, as IsValid(this) would somehow return true - even though that object was deleted completely, going through the UFlowNodeBase destructor.

As flows can survive world changes, the solution is to keep a TWeakObjectPtr<UViewportStatsSubsystem> along with the int32 index returned by the call to UViewportStatsSubsystem::AddDisplayDelegate(). If the UFlowNodeBase is destroyed after a world change, in its BeginDestroy() call, attempt to get the object pointed to TWeakObjectPtr<UViewportStatsSubsystem> - and if it is stale, don't do anything - it got destroyed anyway. Otherwise, remove the display delegate pointed by its index.

The same mechanism is found in UFlowComponent, except that UActorComponent always work in a single world, so no TWeakObjectPtr<> is necessary.

### DIFF
--- a/Source/Flow/Private/FlowComponent.cpp
+++ b/Source/Flow/Private/FlowComponent.cpp
@@ -100,6 +100,28 @@ void UFlowComponent::EndPlay(const EEndPlayReason::Type EndPlayReason)
 	Super::EndPlay(EndPlayReason);
 }
 
+void UFlowComponent::BeginDestroy()
+{
+	// Remark: removing them in the reverse order they where added. It is using RemoveAtSwap()
+	// so removing the first one would change the index of the next one. Removing a delegate in the
+	// middle of the array will invalidate the indices above.
+	if (UWorld* World = GetWorld())
+	{
+		for (int32 Index = ErrorDisplayDelegatesIndices.Num() - 1; Index >= 0; --Index)
+		{
+			if (UViewportStatsSubsystem* StatsSubsystem = World->GetSubsystem<UViewportStatsSubsystem>())
+			{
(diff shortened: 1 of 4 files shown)


MaksymKapelianovych:
> IsValid(this) would somehow return true

We recently had the same mistery on our project - IsValid() returned true, but for null property😐 It disappeared after clean build)

jnucc:
@MaksymKapelianovych  IsValid(UObject*) is ok to use only if non-null, and if the object is still not destroyed (as in pending kill, or marked as garbage but still not collected by the garbage collector). Otherwise after being destroyed, is that memory is not reused it looks like it is still fine, but this can be reused or just completely uncommitted later on and crash (we got the crash on our side because of memory no longer being committed)

MaksymKapelianovych:
> Otherwise after being destroyed, is that memory is not reused it looks like it is still fine

Doesn't the garbage collector automatically reset all UPROPERTY references to nullptr, when an object is deleted?

jnucc:
The changes I've tried here are bad. I have discarded this on our production, and replaced to simply use a TWeakObjectPtr<>.

jnucc:
I'll make a new PR

Issue #277, opened 2025-03-07
Sequence Outputs Other Than 1 & 2 Doesn't Work Anymore


dyanikoglu:
After fetching changes of past 2-3 weeks, sequence nodes stopped working in our project.

![Image](https://github.com/user-attachments/assets/f463f38b-5efa-4d86-89a5-1600b53f7308)

The node looks like this in graph, and the red colored exec outputs doesn't work in runtime. Removing node & readding also doesn't fix the issue.

MaksymKapelianovych:
I also encountered this issue yesterday, it affects not only Sequence node, but any other with manually added pins (MultiGate, etc)

MothDoctor:
It seems to be fixed by @MaksymKapelianovych with this PR :)

PR #279, opened 2025-03-07
Fix issue https://github.com/MothCocoon/FlowGraph/issues/277 by reintroducing some of deleted checks

MaksymKapelianovych:


MaksymKapelianovych:
Reintroduced some checks deleted in https://github.com/MothCocoon/FlowGraph/pull/264, everything is now working fine for me.
FYI @HomerJohnston

MaksymKapelianovych:
Not related to current fix, but do I understand it right, FlowNode can't have both user added and context pins? RefreshContextPins/TryUpdateNodePins will just clear out user added pins without a trace.
It is a rare case, but still a potential bug.

MothDoctor:
> Not related to current fix, but do I understand it right, FlowNode can't have both user added and context pins? RefreshContextPins/TryUpdateNodePins will just clear out user added pins without a trace. It is a rare case, but still a potential bug.

Yes, indeed. I couldn't imagine a case where users might want to add pins while generating them manually. Even if that would be allowed, it's impossible to support every combination of these two approaches. What if user adds pin "X" and afterwards the editor attempts to create context pin "X" based on its data?

Many thanks for the fix :)

HomerJohnston:
Thanks for CC'ing me - I clearly missed this in 264, ugh - however this PR is a bandaid and I think it needs to be fixed further. Basically, with this, I think the code can abort the whole reconstruct node process just because a node does not support context pins. This may give no attention whatsoever to other changes occurring, like data pin changes, or editing native pins from C++ constructor code, or some sorts of other changes.

I'll investigate more today/tomorrow and open a new PR if needed. Flow is getting better. Cheers.

MaksymKapelianovych:
> like data pin changes

Aren't data pins handled in `TryUpdateAutoDataPins()` function?
image

> editing native pins from C++ constructor code

Oh, do you mean hot-reload case? I forgot about it)

HomerJohnston:
OK after a little bit of testing, I think there may not be any notable problems happening after your fix :) so far it seems like it does work fine for auto/data/native/user numbered pins where I thought it might do some bad things, just based on how I was reading the code earlier. 

For context pins, only found one small nuisance so far that requires a manual refresh to fully update the graph node pins which silently go out of sync with the node after you change code. Does not seem like a breaking bug, just a rare edge-case nuisance.

I'll probably keep looking at this whole system over time to see if I can clean it up any more, but I think it's all in a better spot now. Cheers!

PR #281, opened 2025-03-22
Fix copy-paste breaking reroute node directionality

Ryan-DowlingSoka:
image

Because the graph is locked during pasting, the nodes don't reconstruct until the lock finishes, but when the lock does finish the new nodes aren't marked for reconstruction. This fixes the above issue so the pasted nodes look and are properly constructed (with addons too) after pasting.


MothDoctor:
Lovely! Thank you for the fix, accepted :)

PR #282, opened 2025-03-26
Flow debugger update

MaksymKapelianovych:
fixed bugs:
 - fix pausing at breakpoint
 - select correct flow asset instance, instead of a first one spawned with the same name in MP
 - breadcrumbs were not visible if asset editor was opened before PIE start

changes:
 - world selection for inspected instances(hidden if there is only one world). If some world is selected, breakpoints will only be triggered for assets inside that world
 - in the top right corner of the graph PIE status will be displayed, based of selected world/instance
 - if there is an inspected instance of an asset, breakpoints will only be triggered for that instance
 - if there is no inspected instance of an asset, but there are active breakpoints, they will be triggered for all instances of an asset, debugger will temporarily set paused instance as inspected and clear after resume
 - world prefix for instances in dropdown
 - automatic reselection of previously inspected instance between PIE sessions (disabled by default, can be enabled in Editor Settings)
 - visual changes for breadcrumbs (added trailing delimiter, outline, max width, changed text style)
 - clicking on breadcrumb now will focus SubGraph node, that created inspected instance next in chain
 - pausing at breakpoint will focus paused asset window (open if needed) and focus node, that has triggered breakpoint
 - node breakpoint is no longer triggered if there is triggered breakpoint for any pin at the same time
 - "enable/disable/remove all breakpoints" is added to "debug" menu section
 - button "Go to Parent" was removed from the toolbar (the same can be achieved with breadcrumbs), but command was kept in order to let the ability to use shortcut for this action


Source/FlowEditor/Public/Graph/FlowGraphEditorSettings.h
```diff
@@ -69,5 +69,5 @@ class FLOWEDITOR_API UFlowGraphEditorSettings : public UDeveloperSettings
 
 public:
 	virtual FName GetCategoryName() const override { return FName("Flow Graph"); }
-	virtual FText GetSectionText() const override { return INVTEXT("User Settings"); }
+	virtual FText GetSectionText() const override { return INVTEXT("Graph User Settings"); }
```
MaksymKapelianovych: Renamed to avoid name collision with UFlowUserSettings. It would be great if somebody had a better choice for name.

Source/Flow/Public/FlowUserSettings.h
```diff
@@ -0,0 +1,29 @@
...
+#pragma once
+
+#include "Engine/DeveloperSettings.h"
+#include "FlowUserSettings.generated.h"
+
+/**
+ * 
+ */
+UCLASS(Config = EditorPerProjectUserSettings, meta = (DisplayName = "Flow"))
+class FLOW_API UFlowUserSettings : public UDeveloperSettings
```
MaksymKapelianovych: Class for user settings, that can't be moved to Editor module for now. I would also move here `UFlowSettings::bUseAdaptiveNodeTitles`, but it is beyond the scope of this PR.

Source/FlowEditor/Private/Asset/FlowDebugEditorSubsystem.cpp
```diff
@@ -92,12 +137,57 @@ void UFlowDebugEditorSubsystem::OnEndPIE(const bool bIsSimulating)
 	}
 }
 
-void UFlowDebugEditorSubsystem::PauseSession()
+void UFlowDebugEditorSubsystem::PauseSession(const FGuid& FromNode)
 {
-	if (!GUnrealEd->SetPIEWorldsPaused(true))
```
MaksymKapelianovych, line 97: That's why I personally prefer writing `Cond == false` instead of `!Cond` 🙂

MaksymKapelianovych:
Toolbar changes

Before
image

After
image

Several worlds
image

MaksymKapelianovych:
And several words about a bug when several breakpoints can be triggered in the same frame.

The best way to fix this would be to use `FSlateApplication::EnterDebuggingMode()` when flow breakpoint is hit. This function contains a while loop, that updates only Slate and Renderer, allowing editor UI to work, but for any other system code execution is completely stopped. But for this function to work in Flow debugger at least this change in UE source code is required: inside `FInternalPlayWorldCommandCallbacks::IsStoppedAtBreakpoint_InEngineMode()` change `FKismetDebugUtilities::GetCurrentDebuggingWorld() == nullptr` to 'GEditor->PlayWorld == nullptr'. I tested this change locally - everything is fine, BP breakpoints are working, Blutility breakpoints also working, but there is still a chance I missed something. Potentially, this change will allow us even to implement stepping for Flow debugger, if this ever be needed.

Other way it can be fixed is to implement "pin execution request" system. FlowAssets will no longer directly trigger input/output pins in Editor, but direct them to debugger. If there is no hit breakpoints - execute requested pins normally, otherwise store request in pending list, that will be executed after resume. I don't really like this approach, I think it potentially can introduce some bugs because of the postponed execution of nodes.

MaksymKapelianovych:
Important note about world names.

As can be seen on image above, client indexes start with 1 `(Client  1, Client 2, ...)`, when debugging BP they start with 0 `(Client 0, Client 1, ...)`, but on clients' window titles they start with 1 for some reason. So there is inconsistency in UE, and I chose to align with window titles, cause sometimes, after long debugging sessions it is very easy to get confused and forget that debugger indexes are shifted by 1, and spend a long time, trying to debug wrong object...

But I can change it to align with BP debugger, if needed

MaksymKapelianovych:
We can also move breadcrumbs from toolbar to graph tab, like this. Tbh I prefer this layout

image

Source/Flow/Public/FlowUserSettings.h
MothDoctor: Won't be happening for now, as I'm removing the `bKeepLastInspectedInstance` from this PR. I checked this and found that it is quickly confusing, potentially unsafe, and generally a bit of bad practice to cache instance names. Names of instances and the number of them can differ between subsequent runs. I would look in the future to simply distinguish between server instances and clients.

Source/FlowEditor/Public/Graph/FlowGraphEditorSettings.h
MothDoctor: Not needed as `UFlowUserSettings` won't be accepted from this PR.

Source/Flow/Public/FlowUserSettings.h
MothDoctor: The confusing part is where we test multiplayer in the "Standalone" variant. It's clear with the "Listen Server" variant.
image

MothDoctor:
I can finally work on these. There are many useful changes here that I intend to integrate as follows:
- Opening the Flow editor on a breakpoint.
- Not triggering the node breakpoint after we already got the pin breakpoint triggered.
- Additional toolbar to select context. It initially supports worlds (with different `ENetMode` when using online multiplayer), but other contexts might be added in the future (i.e,. `ULocalPlayer` as context while playing local splitscreen).
- "enable/disable/remove all breakpoints" is added to the "debug" menu section. Likely with small code tweaks.

What won't be integrated or has been reworked.
- The debugger will always pause the session on triggering a breakpoint. This is expected behavior, consistent with code breakpoints. Otherwise, users would miss breakpoints because of which instance they checked recently. In the future, we need to find a way to configure conditional breakpoints. Users should be able to choose one breakpoint to only trigger on the server, but another breakpoint in the same graph to only trigger on clients.
- There was an overly complicated logic of retrieving "debug worlds" and labels for them. Lists of worlds and instances were updated with the opening dropdown. And there was way too much iterating on worlds, constructing FString/FText too many times. It's a good practice to keep separated: gameplay/user data (like currently selected world) from the actual user interface.

Review and rework is mostly done, should finish that soon.

PR #284, opened 2025-03-28
Add menu and toolbar extensibility managers.

MaksymKapelianovych:
This will allow to add new options to menu and toolbar for custom flow asset types without the need to extend/override FFlowAssetEditor.


MothDoctor:
Oh, nice, I thought that bit of code would be useless. Good to know I was wrong, change accepted :)

PR #285, opened 2025-03-31
Fix issues with FlowComponent IdentityTags replication.

MichaelCenger:
Fixes:
- Changes to identity tags while offline (NM_Standalone) did not replicate to Clients if going online at some later point.
- Changes to identity tags on the server before BeginPlay was called on the FlowComponent would never replicate to clients
- Multiple calls to UFlowComponent::AddIdentityTag(s) or UFlowComponent::RemoveIdentityTag(s) within a single net update did not get replicated properly. It only would replicate the last of each respective operation.


MothDoctor:
Excellent, thank you for the improvements! :)

MichaelCenger:
You're welcome!
If there are any issues or questions with it at some point, feel free to reach out to me via Github.

PR #286, opened 2025-03-31
ViewportStatsSubsystem delegates: work with TWeakObjectPtr<> instead …

jnucc:
…of checking for IsValid(this), which will always return nullptr once UFlowComponent/UFlowNodeBase is destroyed.


jnucc:
This one should be safe.

As IsValid(this) has no way to know if it is working with a dangling pointer or not inside the ViewportStatsSubsystem delegates, work with TWeakObjectPtr<> on itself instead, and attempt to Get() it. If it returns nullptr, it has been destroyed/garbage collected.

MothDoctor:
Excellent, thank you! :)
I merged it with a similar fix and accepted :) 

https://github.com/MothCocoon/FlowGraph/pull/287

PR #287, opened 2025-04-03
Prevent crash in permanent error log

Deams51:
We occasionally see LogError crashing due to `this` passing the `IsValid` check, although it is not entirely valid anymore. 
Replaced the raw pointer with a weak object pointer in the lambda to prevent it.


MaksymKapelianovych:
This is basically the same as https://github.com/MothCocoon/FlowGraph/pull/286, but less verbose (👍). 
One thing I would change is not to create weak pointer before creating a lambda (extra cope of weak pointer), but to create weak pointer in lambda capture list. See the https://github.com/MothCocoon/FlowGraph/pull/286.

Deams51:
@MaksymKapelianovych Thanks for pointing it out. Fixed.
I didn't realize there was already another PR for this exact issue. 😅

MothDoctor:
Huh, I accepted this and integrated additional micro optimization from https://github.com/MothCocoon/FlowGraph/pull/286 - no need to construct temporary text, if node pointer is invalid.

Thanks! :)

Issue #276, opened 2025-03-07
Ctrl+C & Ctrl+V Issue With Graph


dyanikoglu:
Clicking on a node and doing ctrl+c & ctrl+v directly doesn't paste the copied node while the original node is still selected. It's needed to click on an empty area in graph for Ctrl+V to work.

MaksymKapelianovych:
This is because after adding add-ons copy-pasting logic slightly changed. Now when you have selected node when pasting, code is cheking if selected node can accept pasted nodes as add-ons. If not - nothing gets pasted. Potential fix for this would be to add check if pasted nodes contains as least one add-on, and only if it is true - consider previous check.

MothDoctor:
Please check a fix #289 :)

PR #289, opened 2025-04-09
https://github.com/MothCocoon/FlowGraph/issues/276 - Ctrl+C & Ctrl+V …

jnucc:
…Issue With Graph

On pasted nodes, if there is no involved add-on, proceed with the pasting, making it work like in Flow 1.6.


MothDoctor:
Accepted, thank you for the change!

PR #290, opened 2025-04-10
Flow Breakpoint fixes

jnucc:
Add CanPlaceBreakpoints() functionality in UFlowGraphNode, return false on UFlowGraphNode_Reroute.

In SFlowGraphEditor, in Breakpoint() functions, add conditions to check if CanPlaceBreakpoint() on the selected node is true. Also, since multiple breakpoints on nodes can be set/unset at the same time, make the checks in these functions too. Since they are in a for loop, don't return immediately at the first element, but check if it makes sense to return true on the first item that returns true.

Fix breakpoints that could erroneously be added on reroute nodes through the context menu, or through F9, while not being visible at all.

In UFlowDebuggerSubsystem::MarkAsHit() functions, check if the NodeBreakpoint is enabled.


MothDoctor:
Huh, that's a curious edge case. Change accepted, thank you! :)

PR #291, opened 2025-04-14
Added check for PinFriendlyName.

MaksymKapelianovych:
This will force node to rebuild its pins if only PinFriendlyName is changed.

### DIFF
--- a/Source/FlowEditor/Private/Graph/Nodes/FlowGraphNode.cpp
+++ b/Source/FlowEditor/Private/Graph/Nodes/FlowGraphNode.cpp
@@ -1688,7 +1688,7 @@ bool CheckPinsMatch(const TArray<UEdGraphPin*>& GraphPins, const TArray<FFlowPin
 	{
 		if (!GraphPins.ContainsByPredicate([&FlowNodePin](const UEdGraphPin* GraphNodePin)
 		{
-			return GraphNodePin->PinName == FlowNodePin.PinName;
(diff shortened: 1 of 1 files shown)


jnucc:
We did a very similar change on our side, so the need is there !

There is another `CheckPinsMatch(const TArray<UEdGraphPin*>& GraphPins, const TArray<FFlowPin>& NodePins)` below, I wonder if we need to do the same thing inside the ContainsByPredicate() call ?

MaksymKapelianovych:
Interesting question, I am not sure) Can you provide, please, a test case if you have any, so I could reproduce this issue on my side? @jnucc 
Because I never encountered any issues with that particular function.
Of course we can add check for PinFriendlyName just in case, it shouldn't make any additional problems.

MaksymKapelianovych:
Actually, now when I am looking at the code, looks like that function is redundant and we can just delete it 🤔
But maybe I am missing something

jnucc:
Nah you can forget about the other function below, I'm not all that familiar with all the inner workings of FlowGraph, please ignore. :-)

MaksymKapelianovych:
No no no, that was actually a very valid point. That function also has to be changed and test case is setting pins with friendly names in node constructor. When I made this PR I didn't notice that TryUpdateNodePins refreshes context pins, but not default pins. There is definitely the issue, so thank you for pointing that out)

PR #292, opened 2025-04-25
Prefer AsyncSaveGameToSlot of the synchronous SaveGameToSlot

j0tt:
Hello! I'm working on a title where we essentially banned the use of the synchronous SaveGameToSlot method, in favor of AsyncSaveGameToSlot, which brought me to modify this code in our "fork" (we're in p4) and I decided to have my hand at giving it back.

I get that this save file would be so miniscule that the unease of "causing a hitch" should basically be zero, but I figure, if I understand the code correctly, the save object is synchronously written to memory and any node acting on it will be operating off the in-memory version, while the save object on disk can be written asynchronously as a fire and forget. One thing  I'm unsure of is if there's a scenario where a BP author tries to call `UGameplayStatics::LoadGameFromSlot` using the "FlowSave" slot name, independently of FlowGraph, directly after a checkpoint node. 🤔 

Anyway, at least here for the discussion :)


OccultGameDev:
Maybe this should be configurable in an INI or in a DevSettings? I can see some cases where Synchronous Saving would be preferred, and having it be something we can configure, even if it defaults to Async, seems a good compromise.

MothDoctor:
Hey, finally got here.
I added a setting in the node class that is exposed to DefaultGame.ini. It defaults to `False` to avoid nasty surprises in projects far in development or already released.

image

PR #294, opened 2025-05-28
Fix Linux packaged build by passing by reference

IRSMsoso:
When packaging a project for Linux, I received the following errors

```
FlowAsset.cpp:227:40: error: loop variable 'DeniedNodeClass' creates a copy from type 'const TSubclassOf<UFlowNodeBase>' [-Werror,-Wrange-loop-construct]
  227 |         for (const TSubclassOf<UFlowNodeBase> DeniedNodeClass : DeniedNodeClasses)
      |                                               ^
```

and

```
FlowAsset.cpp:247:41: error: loop variable 'AllowedNodeClass' creates a copy from type 'const TSubclassOf<UFlowNodeBase>' [-Werror,-Wrange-loop-construct]
  247 |                 for (const TSubclassOf<UFlowNodeBase> AllowedNodeClass : AllowedNodeClasses)
      |                                                       ^
```

To solve this, I changed the loop to use a const reference instead of a const copy. As far as I can tell it doesn't change how it functions. I also changed the function signature to take a const reference while I was at it.

### DIFF
--- a/Source/Flow/Private/FlowAsset.cpp
+++ b/Source/Flow/Private/FlowAsset.cpp
@@ -224,7 +224,7 @@ bool UFlowAsset::CanFlowAssetUseFlowNodeClass(const UClass& FlowNodeClass) const
 
 bool UFlowAsset::IsFlowNodeClassInDeniedClasses(const UClass& FlowNodeClass) const
 {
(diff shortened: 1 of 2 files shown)


MothDoctor:
Thanks! This reminds me to set up Clang compilation test on my local build system.

PR #297, opened 2025-06-10
Several improvements for working with user pins

MaksymKapelianovych:
- any user pin can be deleted (previously, default pins, usually 0,1, could not be deleted)
- after a pin is removed all subsequent pins will be correctly moved (previously, removing any pin from node resulted in deleting last user pin)
- added option to insert new pin before or after selected pin
Other changes:
- added transaction for changing signal mode
- removed redundant pin menu entry (BreakPinLinks)


MothDoctor:
Hey, I finally decided not to accept most of changes here.
- Removing the default pin is restricted by design. Users shouldn't mess with that, as it would lead to worse readability of graphs. This would effort to maintaince of all nodes.
- I like the idea of inserting pins, but... it doesn't seem to be fully safe considering Data Pins, AddOns. This might be revisited in the future, but I'm not sure if this can be done without adding effort to maintenance of all nodes. I don't want users to have to care about adding failsafes like `UFlowNode_LogicalOR::CanUserInsertInput`.

I will take this bugfix, but to be submitted from another changelist.
`added transaction for changing signal mode`

image
image

PR #298, opened 2025-06-15
Flow Node Validation enhancements

Numblaze:
### Blueprint Node Validation Support
 * Added support for node validation via Blueprints using a new K2_ValidateNode() function
 * Blueprints can now log validation messages (errors, warnings, notes) using new LogValidation...() functions exposed on UFlowNodeBase

### Flow Asset Validation Logs All Severities
* Before fix : validation only logged warnings and notes if at least one error was present
* After fix : UFlowAsset validation logic ensures that all node validation messages are logged, regardless of their severity

### Flow Node AddOn Node Validation
* The node validation code has been moved from UFlowNode to UFlowNodeBase to allow validation of Flow Node AddOns
* Flow asset validation also validates AddOns

### Example in a Blueprint Flow Node 
image


Source/Flow/Public/Nodes/FlowNode.h
```diff
@@ -59,10 +59,28 @@ class FLOW_API UFlowNode
...
 #endif
 
+	// Flow Node Validation : blueprint compatibility
+
+	// Optional validation override for Blueprints
+	UFUNCTION(BlueprintImplementableEvent, Category = "FlowNode|Validation", meta = (DisplayName = "Validate Node"))
+	EDataValidationResult K2_ValidateNode();
+
+	// Log validation error (editor-only)
+	UFUNCTION(BlueprintCallable, Category = "FlowNode|Validation")
```
MaksymKapelianovych: Maybe add `meta = (DevelopmentOnly)` to these functions?

MaksymKapelianovych:
That's a nice addition for BP nodes, but I noticed that this is for UFlowNode (as previous validation logic). I propose to move it to UFlowNodeBase, it will give users ability to implement validation logic for AddOns also, because currently we can't do that, only runtime message logging is supported.

Source/Flow/Public/Nodes/FlowNode.h
Numblaze: Done !

Numblaze:
> That's a nice addition for BP nodes, but I noticed that this is for UFlowNode (as previous validation logic). I propose to move it to UFlowNodeBase, it will give users ability to implement validation logic for AddOns also, because currently we can't do that, only runtime message logging is supported.

I've taken the feedback into account, and AddOns validation is now possible :)
However, I have an unresolved issue at this stage : when you add a validation log and pass the AddOn reference in the message, the clickable link doesn't redirect to the node in the graph.

Do you have any ideas on the ideal way to do this?
I thought about referencing the Flow Node owner, but it's only valued at runtime in the AddOn.

MaksymKapelianovych:
Unfortunately, SGraphPanel does not know anything about AddOns. When SGraphPanel::Update is called, it loops over top level nodes only, creates their respective SWidgets and stores them in a map NodeToWidgetLookup. When you want to select some node, JumpToNode is called and, eventually, in SNodePanel::Tick it tries to find respective SWidget for UEdGraphNode in NodeToWidgetLookup map (in our case, this is AddOn node) and it is not there :)

We could try to add UEdGraphNodes created for AddOns to Nodes array in UEdGraph, but I don't know if this will work) Simpler way would be just to pass owning FlowNode, but as you said, it is available at runtime only.

image
image

MaksymKapelianovych:
I think we can add `ParentNode` property to AddOns (it will contain pointer to owning `UFlowNode`). We can set it in `UFlowGraphNode::AddSubNode` or `UFlowGraphNode::SetParentNodeForSubNode` (currently only graph node knows about its parent, but it is direct parent, so it can be another AddOn). Then we have to have getter function for `ParentNode` in `UFlowNodeBase`, that will return this for `UFlowNode` and actual parent `UFlowNode` for `UFlowNodeAddOn`. And change this
```
FFlowGraphToken::FFlowGraphToken(const UFlowNodeBase* InFlowNodeBase)
	: GraphNode(InFlowNodeBase->GetGraphNode())
{
	CachedText = InFlowNodeBase->GetNodeTitle();
}
```
to this
```
FFlowGraphToken::FFlowGraphToken(const UFlowNodeBase* InFlowNodeBase)
	: GraphNode(InFlowNodeBase->GetParentNode()->GetGraphNode())
{
	CachedText = InFlowNodeBase->GetNodeTitle();
}

```

Numblaze:
> I think we can add `ParentNode` property to AddOns (it will contain pointer to owning `UFlowNode`). We can set it in `UFlowGraphNode::AddSubNode` or `UFlowGraphNode::SetParentNodeForSubNode` (currently only graph node knows about its parent, but it is direct parent, so it can be another AddOn). Then we have to have getter function for `ParentNode` in `UFlowNodeBase`, that will return this for `UFlowNode` and actual parent `UFlowNode` for `UFlowNodeAddOn`. And change this
> 
> ```
> FFlowGraphToken::FFlowGraphToken(const UFlowNodeBase* InFlowNodeBase)
> 	: GraphNode(InFlowNodeBase->GetGraphNode())
> {
> 	CachedText = InFlowNodeBase->GetNodeTitle();
> }
> ```
> 
> to this
> 
> ```
> FFlowGraphToken::FFlowGraphToken(const UFlowNodeBase* InFlowNodeBase)
> 	: GraphNode(InFlowNodeBase->GetParentNode()->GetGraphNode())
> {
> 	CachedText = InFlowNodeBase->GetNodeTitle();
> }
> ```

I've finally fixed this as you suggested !

MothDoctor:
Hey, I finally managed to review it and merge it!
I handled integrating the changes done after you submitted this PR.

I decided to revert part of the change (caching the owner of AddOn) as it might require more work: this pointer might need to be updated during various editor operations. For now, it's safer to get this pointer dynamically. We might revisit that in the future.

Thank you for the improvements :)

PR #302, opened 2025-06-25
Updates from Riot version q2 2025

LindyHopperGT:
- Adjusted how we subscribe to external UFlowGraphNode changes
- Renamed FFlowNamedDataPinOutputPropertyCustomization (which is now also used for input pins)
- Added enum-based constructors for FFlowDataPinResult* structs
- Fixed InstancedStruct.h reference for UE 5.4
- Added FLOW_API to GetFlowPinType() functions to allow it being called from extension plugins (eg, AIFlowGraph)
- Added FDataValidationContext mutable param to ValidateNode() functions, to allow it to interoperate better with newer-style validation functions, which take the parameter.
- Renamed the parameterless ValidateNode() function to DEPRECATED_ValidateNode().  This will break compiles until the name is changed OR the parameter is added and the ValidateNode() function is updated to use it
- Updated all of the standard flow nodes' ValidateNode functions to use the context parameter.
- Created a new UFlowNode_FormatText that formats text using input pins and FText formatting engine
- Changed the UFlowNode_Log to format text (a la UFlowNode_FormatText) to generate its logged output string.
- Changed UFlowNode_Log to inherit from UFlowNode_DefineProperties, so that it can have input properties added on the instance
- Renamed UFlowNode_DefineProperties::OutputProperties to NamedProperties, so that it can be used as the super class for UFlowNode_FormatText
- Added GetRandomSeed() function to UFlowNode.  The default version uses the hash from the node's GUID.  This can be overridden in subclasses (which we do) to any implementation that suits the client code.
- Added an IsFinishedState() classifier function for EFlowNodeState, to error-proof checking node state for 'finished' states
- Updated the ForEachAddOn templates to support a parameter to control how the function should recurse into child addons (or not).
- Fixed UFlowNode_ExecuteComponent to handle injected components correctly in validation
- Fixed UFlowNode_ExecuteComponent to conform to the new style of pin generation, now using ContextPins (the old method didn't work after a refactor with flow graph node reconstruction)
- Updated UFlowNode_ExecuteComponent to allow the component to supply data pin output values
- Updated the networking of runtime IdentityTag changes
- Added UFlowAsset::GatherNodesConnectedToAllInputs helper function
- Extracted UFlowNodeAddOn::FindOwningFlowNode() functionality into a function


MaksymKapelianovych:
Hey. I had a quick check and looks like you did not do a proper merge with other PRs, as many of them since April (#294, #291, #287, #286, #285) have been reverted. Please, bring them back or, if this was done for some reason, explain why)

Also, you changes to validation logic contradict with this PR #298, which which brings FlowNode's validation to BP. To achieve the same result with your changes we would need to still keep custom validation log for BP, managing 2 ways of doing the same thing. Maybe it would be better to keep current approach?

LindyHopperGT:
I’ll go through those PRs and reintegrate them. Sometimes it’s hard to tell which direction a change is coming from (our side vs yours)Cheers,Greg TaylorOn Jun 26, 2025, at 7:24 PM, MaksymKapelianovych ***@***.***> wrote:﻿MaksymKapelianovych left a comment (MothCocoon/FlowGraph#302)
Hey. I had a quick check and looks like you did not do a proper merge with other PRs, as many of them since April (#294, #291, #287, #286, #285) have been reverted. Please, bring them back or, if this was done for some reason, explain why)
Also, you changes to validation logic contradict with this PR #298, which which brings FlowNode's validation to BP. To achieve the same result with your changes we would need to still keep custom validation log for BP, managing 2 ways of doing the same thing. Maybe it would be better to keep current approach?

PR #304, opened 2025-07-20
fix some missing includes

bohdon:
Hello! I ran into a few compiler errors after adding the plugin (v2.1-5.6 / c4c33f5) to a fresh, stock, 5.6 project. Seems like they are hidden by unity-compilation perhaps?

To test, I added this to all 3 Build.cs files of the plugin to very unity-compilation isn't hiding any others (not part of this PR):
```
bUseUnity = false;
PCHUsage = PCHUsageMode.NoPCHs;
```

The `FlowGraphSchema.h` one is tricky, without that include the `ENGINE_MAJOR_VERSION`/`ENGINE_MINOR_VERSION` aren't defined so we get the definition of a function in the .cpp without the declaration in this .h

### DIFF
--- a/Source/FlowEditor/Public/Asset/SFlowDiff.h
+++ b/Source/FlowEditor/Public/Asset/SFlowDiff.h
@@ -4,6 +4,7 @@
 
 #include "IDetailsView.h"
 #include "DiffResults.h"
+#include "GraphEditor.h"
 #include "SDetailsDiff.h"
 #include "Textures/SlateIcon.h"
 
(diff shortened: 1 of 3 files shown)


MothDoctor:
Thanks, I need to look into these settings again. Not sure if I want to disable this compilation optimization by default, but perhaps there's a way to disable on build machines only, will check it :)

bohdon:
Yeah I definitely only do it temporarily to clean things up occasionally 👌🏼

PR #305, opened 2025-07-25
Updates from Riot version end of Q2 2025

LindyHopperGT:
Renamed FFlowNamedDataPinOutputPropertyCustomization (which is now also used for input pins)
Added enum-based constructors for FFlowDataPinResult* structs
Fixed InstancedStruct.h reference for UE 5.4
Added FLOW_API to GetFlowPinType() functions to allow it being called from extension plugins (eg, AIFlowGraph)
Created a new UFlowNode_FormatText that formats text using input pins and FText formatting engine
Changed the UFlowNode_Log to format text (a la UFlowNode_FormatText) to generate its logged output string.
Changed UFlowNode_Log to inherit from UFlowNode_DefineProperties, so that it can have input properties added on the instance
Renamed UFlowNode_DefineProperties::OutputProperties to NamedProperties, so that it can be used as the super class for UFlowNode_FormatText
Added GetRandomSeed() function to UFlowNode. The default version uses the hash from the node's GUID. This can be overridden in subclasses (which we do) to any implementation that suits the client code.
Added an IsFinishedState() classifier function for EFlowNodeState, to error-proof checking node state for 'finished' states
Updated the ForEachAddOn templates to support a parameter to control how the function should recurse into child addons (or not).
Fixed UFlowNode_ExecuteComponent to handle injected components correctly in validation
Fixed UFlowNode_ExecuteComponent to conform to the new style of pin generation, now using ContextPins (the old method didn't work after a refactor with flow graph node reconstruction)
Updated UFlowNode_ExecuteComponent to allow the component to supply data pin output values
Added UFlowAsset::GatherNodesConnectedToAllInputs helper function
Extracted UFlowNodeAddOn::FindOwningFlowNode() functionality into a function


MothDoctor:
Hey, thanks again for a lot of improvements and fixes!

However, I need ask if it would be possible to break future changes into smaller chunks? It should be quite convenient to do, if using a client like GitHub Desktop or similar :)

LindyHopperGT:
yeah, this was a bunch at once, sorry about that.  we had a lockdown for a
deliverable so I didn't get to merge with flow for ~6 months.  it all piled
up.  We're back to more regular updates now.  I think we're on 5.5 now, and
merging 5.6 currently.

Thanks for taking the updates.  Will have some search improvements
hopefully in the next six months or so.  we have it on the roadmap at least.

MothDoctor:
> I think we're on 5.5 now, and merging 5.6 currently. 
Good to know, I will be free now to remove UE 5.4 support ;)

PR #306, opened 2025-08-08
Fix the Datapin result logic

DemonViglu:
When I used the data pin without connecting to any FlowNode, its datapinResult returned Success which may has a fault? According to the code, when the pin's prerequisties is correct and couldn't get any SupplierData means FailedUnconnect. So I guess maybe the code to change the Result  should in the condition : failed to TryGetFlowDataPinSupplierDatasForPinName. Otherwise, TrySetupWorkingData is success but result is fail doesn't make sense;


MothDoctor:
Thank you for your contribution, merged in! :)

PR #308, opened 2025-10-03
Flow graph flow asset params initial

LindyHopperGT:
[FlowAssetParams] Initial implementation of FlowAssetParams
FlowAssetParams are a new data asset, that is optionally managed by the FlowAsset, sourcing properties from the Start node of the graph.
These params can be subclassed to create overrides.  At runtime, either the base or a compatible subclass may be used to provide data for start data pins in the flow graph.

- Added FlowAssetParams
- Added support to FlowAsset to optionally create and sync with FlowAssetParams base
- Added AssetDefinition_FlorAssetParams to add 'create child', for creating 'subclassed' params that override some elements
- Updated FFlowNamedDataPinProperty to have Guid and other properties used in the FlowAssetParams use-case
- Added FlowNamedPropertiesSupplierInterface
- Added FlowAssetProviderInterface
- FFlowAssetParamsPtr - wrapper class for details customization
- Added Params support to FlowComponent & FlowSubsystem

Also:
- Removed stale references to IFlowOwnerInterface
- Merged over some deltas from IFlowExtended/Curated Property Customization


Source/Flow/Public/Asset/FlowAssetParamsUtils.h
```diff
@@ -0,0 +1,44 @@
...
+#include "FlowAssetParamsUtils.generated.h"
+
+class UObject;
+struct FFlowNamedDataPinProperty;
+
+/**
+* Utility functions for Flow Asset Params reconciliation and validation.
+*/
+USTRUCT()
+struct FLOW_API FFlowAssetParamsUtils
```
gregorhcs, line 17: I'd consider using a namespace here: they tend to make the intent for such utility function collections clearer and eliminate USTRUCT compile overhead

MothDoctor:
Closing this as this was only a "preview" ahead of the big data pins refactor.

PR #309, opened 2025-10-06
Enhance validation in `FlowGraphNode` by adding null checks

Numblaze:
The initial issue I wanted to fix was the Ctrl + Shift + X shortcut in the BlueprintAssist plugin, which was causing a crash. This shortcut deletes the selected nodes on the graph and rebuilds the execution pin as if they didn't exist.


MothDoctor:
Thank you for your contribution, merged in! :)

PR #312, opened 2025-10-18
Fixed: build error when making installed build

ameaninglessname:


MothDoctor:
Change integrated. Thank you! :)

Issue #314, opened 2025-12-13
OnNotifyFromActor - Ignores Identity tags match type


SilentGodot:
The node `OnNotifyFromActor` inherits the identity tag match type from the ObserveActor node, however in the function `OnNotifyFromComponent` it strictly calls `HasAnyExact` on the identity tags and doesn't use the match type.

<img width="780" height="399" alt="Image" src="https://github.com/user-attachments/assets/dd4ee0eb-d95a-4eb0-bc83-80911e5bc841" />

```
void UFlowNode_OnNotifyFromActor::OnNotifyFromComponent(UFlowComponent* Component, const FGameplayTag& Tag)
{
	if (Component->IdentityTags.HasAnyExact(IdentityTags) && (!NotifyTags.IsValid() || NotifyTags.HasTagExact(Tag)))
	{
		OnEventReceived();
	}
}
```

This means any child tags will not be matched. In the attached screenshot for example, any child tag of `Flow.Pawn.Collectible.Tombstone` won't trigger the notify.

SilentGodot:
Uploaded a suggested fix here: https://github.com/MothCocoon/FlowGraph/pull/315

MothDoctor:
Sugessted fix testes and accepted :)

PR #315, opened 2025-12-13
Fix OnComponentNotify node not using identity tag match type

SilentGodot:
Fix issue https://github.com/MothCocoon/FlowGraph/issues/314#issue-3725743097

The node OnNotifyFromActor inherits the identity tag match type from the ObserveActor node, however in the function OnNotifyFromComponent it strictly calls HasAnyExact on the identity tags and doesn't use the match type. 
This means any child tags will not be matched even if the match type is set to non-exact.


MothDoctor:
Thanks for the fix! It was a silly oversight, indeed ;)

PR #316, opened 2025-12-14
Add "Run from this node" option to Flow Graph context menu

kathelon:
- Added a new context menu action to start execution from a selected node during PIE.
- Updated FlowAsset and FlowSubsystem to support starting from a specific node GUID.

resim


MothDoctor:
I decided not to accept these changes, for a few reasons.
* We already have the `Force Pin Activation` feature. Any related feature shouldn't integrate with the existing code path and UI.
https://mothcocoon.github.io/FlowGraph/Features/ForcePinActivation
* Bypassing the graph signal entirely would create issues due to instantiating the graph without the context of other graphs. Data Pins might not be correctly initialized.
* Any other work touching the runtime execution flow of the graph would need to take this specific feature into account.

I would suggest moving this feature to the project-specific code by extending base plugin classes. This way your team can preserve this feature without the need for integrating the plugin's code :)

PR #319, opened 2025-12-25
Exposed methods to public to enable flow node status display in runtime debuggers

gregorhcs:
We've been a building a gameplay debugger integration, and for that it's neat to see the active nodes' statuses. This change enables that by moving status-fetchers out of editor-only clauses.


MothDoctor:
Hey, I don't have rights to update your branch, so I implemented it as a might changelist. It was actually the most convenient way due to many changes made by recently approved PRs.

PR #320, opened 2025-12-25
Added virtual method to allow saving non-executing flow nodes

gregorhcs:
In our project there were cases, where data pin wiring between nodes led to corrupt state after loading, because already finished nodes wouldn't restore their state, state that sourced their outgoing data pins which were used by now active nodes.

Thus we added a way to save nodes despite being already executed, which we enabled on all nodes that provided critical data pin values.


MothDoctor:
Oh, this is a pretty interesting case. It is elegant way of handling that, so accepted :)

PR #321, opened 2025-12-25
Made IsBoundToWorld() const

gregorhcs:
This small getter shouldn't need to ever modify state.


MothDoctor:
You're most likely correct. Change integrated :)

PR #322, opened 2025-12-25
Exposed flow subsystem runtime state fields to subclasses

gregorhcs:
For our custom flow subsystem, we needed access to this state; shouldn't have a big impact.


MothDoctor:
Sure, integrated :)

PR #323, opened 2025-12-25
Fixed rare crash in LogError(..) caused by invalid flow node self or owner

gregorhcs:


MothDoctor:
Curious. Thanks for adding a nullcheck :)

PR #324, opened 2025-12-25
Added option to tone down on-screen error messages

gregorhcs:
In our project, we try to avoid on screen messages altogether since they're popping up for people that are not the target audience. This PR adds an option to tone down the error message for finished/aborted nodes being triggered, as well as a general error log option to not show the message on screen.

I'm open to suggestions how to maybe widen the coverage of any setting toning down screen errors - maybe such an option should cover all log error cases?


MothDoctor:
Hey, it's indeed a good idea to provide a way to prevent throwing these messages on the screen.
Only it seemed a bit clunky to create a new Project Settings entry for a single log. I was considering replacing that with cvar, but then I was like, "wait, we can actually disable throwing this to the screen altogether"? I guess I was too aggressive in throwing messages on the screen, so I disabled that for everyone. Simple.

Thank you for this change :)

PR #326, opened 2026-01-12
Flow Editor Search Improvements

LindyHopperGT:
- Updated Search UX, displays the source(s) where the search terms were found
- Multi-asset search capability (same time or all flow assets)
- Increased number of search sources (Eg, tooltips or property values)
- Filters to control which search sources to include
- Improved inline object, struct and subgraph search capabilities
- Caching search meta-data for each flowasset, for faster re-searches, updated when the asset changes

### DIFF
--- a/Source/Flow/Private/Nodes/Graph/FlowNode_FormatText.cpp
+++ b/Source/Flow/Private/Nodes/Graph/FlowNode_FormatText.cpp
@@ -7,6 +7,8 @@
 
 #define LOCTEXT_NAMESPACE "FlowNode_FormatText"
 
+const FName UFlowNode_FormatText::OUTPIN_TextOutput("Formatted Text");
+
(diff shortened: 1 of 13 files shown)


Source/Flow/Private/Nodes/Graph/FlowNode_FormatText.cpp
```diff
@@ -7,6 +7,8 @@
 
 #define LOCTEXT_NAMESPACE "FlowNode_FormatText"
 
+const FName UFlowNode_FormatText::OUTPIN_TextOutput("Formatted Text");
```
LindyHopperGT: This was removed in the merge for Data Pins Update, and I disagree with the inlining of the text -- it gives more opportunities for mistyping and having the link break because someone didn't update every reference text.

PR #328, opened 2026-01-26
Flow debugger fixes

LindyHopperGT:
(Branch from the Search PR branch)

- Data Pin values visible in the debugger as tooltips
- Debugger will stop at breakpoints instead of continuing to trigger


MothDoctor:
Thank your changes. I did integrate that with the previous breakpoint improvements.
https://github.com/MothCocoon/FlowGraph/commit/a3318d01e402741f256252f513a84f63b41e3370

PR #333, opened 2026-02-05
Flow deferred trigger queuing

LindyHopperGT:
Trigger Outputs deferred while processing an Input Trigger
This is a change to the core flow triggering logic to fix a category of sequencing bugs from the previous behavior.  It would immediately fully process a triggered input and so on down the chain of flow nodes, without allowing the current flow node to finish executing, this caused a whole category of problems where the node wasn't able to finish its execution before being interrupted by a retirgger (from downstream) and AddOns wouldn't execute at the same time as their owning flow node reliably
Now FlowAsset will queue any triggers generated while processing a trigger, and flush them when ending the processing of that trigger.
We also integrated the debugger queued trigger caching mechanism to use the same syste.
Subclasses of UFlowAsset that do their own deferred asset triggering can disable this feature, except for the debugger portion, which is still processed using the UFlowAsset queue.

CR (Riot) - JDurica, BJarvinen


MothDoctor:

PR #334, opened 2026-02-05
Flow data pin support for add ons

LindyHopperGT:
- AddOns can now include data pins, which show up as pins on their owning node
- Duplicate outputs are disambiguated with different names
- UFlowNode has a map for non-trivial data pins  (trivial are those which exist in the 0th object source and have a PinName that exactly matches their source PropertyName)

CR (Riot) - JDuriva, BJarvinen


MothDoctor:

PR #335, opened 2026-02-06
NamedReroute Nodes

C209:
Add NamedReroute Nodes for better graph management.

### DIFF
--- a/Source/Flow/Private/FlowAsset.cpp
+++ b/Source/Flow/Private/FlowAsset.cpp
@@ -13,6 +13,7 @@
 #include "Nodes/Graph/FlowNode_CustomOutput.h"
 #include "Nodes/Graph/FlowNode_Start.h"
 #include "Nodes/Graph/FlowNode_SubGraph.h"
+#include "Nodes/Route/FlowNode_NamedReroute.h"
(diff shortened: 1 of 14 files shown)


MothDoctor:
I'm gonna be honest with you. I don't think I will decide to integrate this into the mainline branch. The fundamental design of the Flow Graph is to keep graphs super-clear. Easy to read, debug, and maintain. This was a conscious decision to enforce having visible links between all nodes. Without that, approaches like "debug by wire" can't be fully utilized.

I'm open to further discussion here and on the plugin's Discord.
I don't close PR, as we can safely assume that quite a few teams would actually ignore the design approach described above. And they would be happy to integrate Named Reroute to their projects :)

MothDoctor:
Hey @C209, I'm closing this PR. I've talked with community and the consensus confirms my original design: Named Reroutes would break the readability and simplicity of Flow Graph. And the same this would increase the maintenance cost of the system.

But it's clear some teams would like to have Named Reroutes. They would like to know that such a modification exists. So I added a documentation page: Notable Pull Request, and linked to this sumbit there :)
https://mothcocoon.github.io/FlowGraph/Forks/NotablePullRequests

C209:
I totally understand your design. It's good to leave the implementation here for the teams that need it.

PR #336, opened 2026-02-06
Flow data pin reroute support

LindyHopperGT:
- Added Reroute node support for Flow Data Pins (previously was Exec pins only)
- Slight cost during lookup due to passing through the reroute node(s) to get to the actual node 
- Same cost as regular reroutes in the form of full flow node for a clerical purpose.  (It would be nice to strip these out of the graph for shipping flow graphs, but that's a future goal)

CR (riot) - JDurica, BJarvinen


MothDoctor:

PR #337, opened 2026-02-11
Flow asset params improvements

LindyHopperGT:
**[Flow] Converted FlowDataPinValueSupplierInterface to C++ only**
The blueprint functionality for this has been removed
Rationale:
The Execute_TrySupplyDataPin overhead is non-trivial vs. a simple virtual call in C++
Could still create custom supply logic in C++
It's a really technical thing to do in Blueprint
The new Data Pin Value wrappers for blueprint are highly functional and usable; providing a very clean and easy method to add data pins to blueprint nodes
Removing the support could allow me to simplify the Pin authoring interface in blueprint (for exec pins) to make it more clear, since they wouldn't need to also support authoring data pins in that way.

**[Flow] Added FlowNode_BlueprintDataPinSupplierBase**
This is a thin c++ class that exposes TrySupplyDataPin override capability for blueprint if they need it.
Suggested by a thread on the Flow Discord, where a user indicated a legitimate use case for still wanting to override TrySupplyDataPin in blueprint
This class provides a targeted override capability to blueprint, while still leaving the blueprint dispatch out of the UFlowNode base implementation (for better perf in the general case)
CR - BJarvinen, ECalder

**[Flow] Added factory creation method for FlowAssetParams**
Previously, we only supported flow asset params by right-clicking the desired parent asset, now we also have a Create Asset entry (like Flow Asset, or Flow Node Blueprint, etc.)
The new factory create still requires a parent params asset, which is selected via an asset-picker.
Created common child asset creation code for both methods of creating child FlowAssetParams to share
CR - BJarvinen

**[Flow] FlowAssetParams configuration for Subgraph nodes**
Can now statically assign AssetParams to use on a Subgraph node for the associated (flow) Asset
Can also dynamically source the AssetParams to use via an input data pin for the Subgraph node
If the Subgraph node pins are connected, the connected supplier is preferred over the FlowAssetParams
Otherwise, the FlowAssetParams object (if chosen) will supply the value (this is the new part)
If no assigned FlowAssetParams, the Start node in the subgraph's default values are used (this was pre-existing behavior)
Also:
Fixed some code in and around sourcing subgraph Start node pin values to allow this sourcing to work properly
Fixed some issues preventing the tooltip inspection for data pins in subgraphs from correctly finding the correct source value
Removed an incorrect IsHalted check, which was preventing subgraphs from starting when at a breakpoint (as they aren't queued, this would just drop the start of the subgraph rather than deferring it)
CR - BJarvinen, ECalder

**[Flow] Data Pin Reroute support**
Added Reroute node support for Flow Data Pins (previously was Exec pins only)
Slight cost during lookup due to passing through the reroute node(s) to get to the actual node
Same cost as regular reroutes in the form of full flow node for a clerical purpose.  (It would be nice to strip these out of the graph for shipping flow graphs, but that's a future goal)
CR - JDurica, BJarvinen


MothDoctor:

PR #338, opened 2026-02-12
Fix FlowNode_FormatText's pin direction

LindyHopperGT:
- All of the NamedProperty pins were output, but should have been input
- Adapted the FlowNode_Log code to fix the pin direction to be sharable between these two nodes

CR - BJarvinen, ECalder


MothDoctor:

PR #345, opened 2026-02-17
Created "Switch" node for Flow

LindyHopperGT:
**[Flow] Updated FlowNode_Branch: AND/OR selectability and BranchCase support**
Branch node can now be configured with AND/OR for the top-level combination rule (default is still AND)
Branch node can now have BranchCase AddOns at its root addon level, these will be evaluated and can trigger prior to evaluating the root predicates (which serve as the "else" case for this sort of configuration)
BranchCase adds a switch/case-like capability where each BranchCase can have its own output exec pin, if that case's predicates pass
CR - BJarvinen, ECalder

**[Flow] Switch Case AddOns' names are authorable**
If you set the name, it will match the pin 1:1 (unless duplicates, see below)
Default case name (and pin name) is "Case".
Duplicates are disambiguated with numbers appended.
Sets the title to the pin name, so it's easier to match the pin to the case
CR - JDurica, BJarvinen


MothDoctor:
Closing as duplicate of https://github.com/MothCocoon/FlowGraph/pull/346

PR #352, opened 2026-03-03
Compare Values Predicate

LindyHopperGT:
[Flow] Create CompareValues predicate (for data pins) and auto-generate data pins refactor

- Refactored the auto-generate data pins code so that the CompareValues predicate can get its pins generated, duplicates disambiguated and the results queried
- Created CompareValues predicate, which is analogous to the Compare Blackboard Values predicate, but for data pins

CR - JDurica, BJarvinen, ECalder


Source/Flow/Private/AddOns/FlowNodeAddOn_PredicateCompareValues.cpp
```diff
@@ -0,0 +1,605 @@
...
+
+	return CompareDoubleUsingOperator(LeftDouble, RightDouble);
+}
+
+bool UFlowNodeAddOn_PredicateCompareValues::TryCompareAsInt64() const
+{
+	int64 LeftInt64 = 0;
+	{
+		const EFlowDataPinResolveResult ResolveResult =
+			TryResolveDataPinValue<FFlowPinType_Int64>(GetDisambiguatedValueName(LeftValue), LeftInt64, SingleFromArray);
```
LindyHopperGT: These resolves using the "disambiguated name" is notable. This name is the version the owning flow node uses to identify this pin. We need to ask for that name specifically, so that it routes back to our addon for the default value (for an unconnected pin) and also so that the owning flow node can find the pin in its list of pins (as it wouldn't be findable under the authored name, in the case of duplicate pin names).

Source/Flow/Private/Interfaces/FlowDataPinValueOwnerInterface.cpp
```diff
@@ -0,0 +1,73 @@
...
+	const FProperty*& OutFoundProperty,
+	TInstancedStruct<FFlowDataPinValue>& OutFoundInstancedStruct) const
+{
+	const UObject* ThisAsObject = Cast<UObject>(this);
+	return TryFindPropertyByPinName_Static(*ThisAsObject, PinName, OutFoundProperty, OutFoundInstancedStruct);
+}
+
+bool IFlowDataPinValueOwnerInterface::TryFindPropertyByPinName_Static(const UObject& PropertyOwnerObject, const FName& PinName, const FProperty*& OutFoundProperty, TInstancedStruct<FFlowDataPinValue>& OutFoundInstancedStruct)
+{
+	// First check if the PinName matches a NamedProperties array name
```
LindyHopperGT: This part is new, we used to have this built-in to the nodes that have NamedProperties arrays. Now those nodes don't need to handle it specially, since the 'standard' implementation handles it now.

Source/Flow/Private/Nodes/FlowNode.cpp
```diff
@@ -427,59 +427,94 @@ void UFlowNode::RemoveUserOutput(const FName& PinName)
 
 bool UFlowNode::TryUpdateAutoDataPins()
 {
-	// Generate data pins, feed list to working data struct
-	FFlowAutoDataPinsWorkingData WorkingData = FFlowAutoDataPinsWorkingData(AutoInputDataPins, AutoOutputDataPins);
-	AutoGenerateDataPins(WorkingData);
+	FFlowAutoDataPinsWorkingData WorkingData(AutoInputDataPins, AutoOutputDataPins);
```
LindyHopperGT: This is a GPT refactor of this function to ensure that we don't mutate the pin properties unless we actually are making a deviation from the current value. All of the actual patching is now deferred (see the end of this function) until after we decide that the pin name changes, etc. are significant and are worth Modify()-ing the object.

This reduces the amount of file-dirtying over no-op changes and it's a safer pattern overall.

Source/Flow/Private/Types/FlowAutoDataPinsWorkingData.cpp
```diff
@@ -57,38 +80,49 @@ void FFlowAutoDataPinsWorkingData::BuildNextFlowPinArray(const TArray<FFlowPinSo
...
+	void* ValueOwnerAsObject = ValueOwner.GetValueOwnerAsObject();
 
-	const FStructProperty* StructProperty = CastField<FStructProperty>(Property);
+	FStructProperty* StructProperty = CastField<FStructProperty>(Property);
 	const FFlowDataPinValue* DataPinValue = nullptr;
 	if (StructProperty && StructProperty->Struct)
 	{
 		const UScriptStruct* ScriptStruct = StructProperty->Struct;
 
+		if (StructProperty->Struct->IsChildOf<FFlowNamedDataPinProperty>())
```
LindyHopperGT: This 'standard' support allowed me to skip a special case in the subclass' pin generation.

If it stumbles upon a FlowNamedDataPinProperty, it will generate the pin for it here. This only applies to the PredicateCompareValues for now, but it seemed better to handle it in standard, than to special case it in that addon.

Source/Flow/Private/Types/FlowNamedDataPinProperty.cpp
```diff
@@ -32,6 +33,25 @@ FText FFlowNamedDataPinProperty::BuildHeaderText() const
...
 }
+
+void FFlowNamedDataPinProperty::AutoGenerateDataPinForProperty(const FFlowDataPinValueOwner& ValueOwner, FFlowAutoDataPinsWorkingData& InOutWorkingData)
+{
+	if (!IsValid())
+	{
+		return;
+	}
+
+	FFlowDataPinValue& Value = DataPinValue.GetMutable();
```
LindyHopperGT: This code used to be inlined in cases where we generated data pins for NamedDataProperties. Now this class knows how to auto-generate its own data pin, in a 'standard' location.

Source/Flow/Private/Types/FlowPinTypesStandard.cpp
```diff
@@ -71,7 +71,8 @@ bool FFlowPinType_Enum::PopulateResult(const UObject& PropertyOwnerObject, const
 	TInstancedStruct<FFlowDataPinValue> ValueStruct;
 	const FProperty* FoundProperty = nullptr;
 
-	if (!Node.TryFindPropertyByPinName(PropertyOwnerObject, PropertyName, FoundProperty, ValueStruct))
+	const IFlowDataPinValueOwnerInterface* PropertyOwnerInterface = CastChecked<IFlowDataPinValueOwnerInterface>(&PropertyOwnerObject);
+	if (!PropertyOwnerInterface->TryFindPropertyByPinName(PropertyName, FoundProperty, ValueStruct))
```
LindyHopperGT: Moved this function to the interface.

Source/Flow/Public/Types/FlowAutoDataPinsWorkingData.h
```diff
@@ -33,43 +114,95 @@ struct FFlowPinSourceData
...
 	FFlowAutoDataPinsWorkingData(const TArray<FFlowPin>& InputPinsPrev, const TArray<FFlowPin>& OutputPinsPrev)
 		: AutoInputDataPinsPrev(InputPinsPrev)
 		, AutoOutputDataPinsPrev(OutputPinsPrev)
 	{
 	}
 
-	FLOW_API bool AutoGenerateDataPinsForFlowNode(UFlowNode& FlowNode, bool& bAutoInputDataPinsChanged, bool& bAutoOutputDataPinsChanged);
-	FLOW_API void AddFlowDataPinsForClassProperties(const UObject& ObjectContainer, int32 PropertyOwnerIndex);
-	FLOW_API static void BuildNextFlowPinArray(const TArray<FFlowPinSourceData>& PinSourceDatas, TArray<FFlowPin>& OutFlowPins);
+	/* Builds the proposed next pins, property source map, and staged wrapper patches. */
```
LindyHopperGT: Reworked all of these functions to make them more clear sequence of events.

PR #357, opened 2026-03-12
Crash fix for Format Text node

LindyHopperGT:
Prevent Getting the value of an invalid FFlowDataPinValue property.


Source/Flow/Private/Nodes/FlowNodeBase.cpp
```diff
@@ -990,6 +990,11 @@ EDataValidationResult UFlowNodeBase::ValidateNode()
 
 bool UFlowNodeBase::TryAddValueToFormatNamedArguments(const FFlowNamedDataPinProperty& NamedDataPinProperty, FFormatNamedArguments& InOutArguments) const
 {
+	if (NamedDataPinProperty.Name.IsNone() || !NamedDataPinProperty.DataPinValue.IsValid())
```
LindyHopperGT, line 993: The code that follows was trying to inspect the invalid FFlowDataPinValue struct.

PR #359, opened 2026-03-12
Bugfixes and refactor for LogWarning/Error/Note

LindyHopperGT:
**Bug Fix	 - CancelAndWarnForUnflushedDeferredTriggers()**
- Null-guard ToNode and FromNode before dereferencing in UE_LOG. Prevents crash during abnormal termination when nodes are already destroyed.

**Bug Fix	 - TryFindActorOwner()**
- Now correctly returns the Owner when it IS already an AActor, not just when it's a component. Fulfills the documented contract.

**Refactor - LogError/LogWarning/LogNote**
- Extracted shared LogRuntimeMessage() helper. Three identical copy-pasted functions → one implementation with severity parameter.

**Policy System (FFlowPinConnectionPolicy)**
UFlowSettings exposes default policies as nullable const T* accessors. UFlowAsset holds a resolved policy instance with check() validation.


Source/Flow/Private/Nodes/FlowNode.cpp
```diff
@@ -1183,16 +1185,161 @@ void UFlowNode::RecursiveFindNodesByClass(UFlowNode* Node, const TSubclassOf<UFl
...
+
+	Super::Cleanup();
+}
+
+void UFlowNode::ExecuteInput(const FName& PinName)
+{
+	// Often ExecuteInput is replaced rather than extended in subclasses.
+	// So any subclasses that implement the preload interface will want to call this function 
+	// in their ExecuteInput() override.
+	if (DispatchExecuteInputToPreloadHelper(PinName))
```
dskliarov-gsc: I have a small suggestion.
1. Declare DispatchExecuteInputToPreloadHelper as a FlowNodeBase virtual method
2. Make it return false in the default implementation
3. Keep the same implementation for FlowNode, like the one you have here
4. Remove the DispatchExecuteInputToPreloadHelper call from the UFlowNode::ExecuteInput
5. Place it under the UFlowNodeBase::ExecuteInputForSelfAndAddOns instead, so we may get something like this

```
void UFlowNodeBase::ExecuteInputForSelfAndAddOns(const FName& PinName)
{
	// AddOns can introduce input pins to Nodes without the Node being aware of the addition.
	// To ensure that Nodes and AddOns only get the input pins signaled that they expect,
	// we are filtering the PinName vs. the expected InputPins before carrying on with the ExecuteInput

	if (IsSupportedInputPinName(PinName) && !DispatchExecuteInputToPreloadHelper(PinName))
	{
		ExecuteInput(PinName);
	}

	for (UFlowNodeAddOn* AddOn : AddOns)
	{
		AddOn->ExecuteInputForSelfAndAddOns(PinName);
	}
}
```

There are two main reasons why I think it may work better:
1. Preload/Flush are not just regular execution pins. We do not expect them to perform any additional logic except for managing the resources. In such a way, it will be easier to prevent human mistakes when we forget to protect the main execution block from being called by Preload/Flush pins
2. As you mentioned here, sometimes ExecuteInput is replaced rather than extended in subclasses. And with this new approach, we don't need to duplicate the DispatchExecuteInputToPreloadHelper for such nodes

What do you think?

PR #361, opened 2026-03-30
Flow Preload and Policy changes

LindyHopperGT:
[Flow] Replaced the stubbed, nonfunctional preload mechanism in FlowGraph with a policy-driven, async-safe & per-project extendable system. Core Interface (IFlowPreloadableInterface)
Replaced ad-hoc PreloadContentAsync (TFunction callback) with PreloadContent() → EFlowPreloadResult (Completed / PreloadInProgress) and FlushContent(). Async C++ nodes return PreloadInProgress and call NotifyPreloadComplete() from their completion delegate. Async Blueprint nodes override K2_PreloadContent and call NotifyPreloadComplete() on self. Sync nodes return Completed unchanged.

Policy System (FFlowPreloadPolicy, FFlowPinConnectionPolicy)
Introduced FFlowPreloadPolicy instanced-struct policy controlling per-node preload timing (OnGraphInitialize, OnActivate, ManualOnly, Never) and flush timing (OnGraphDeinitialize, OnNodeFinish, ManualOnly, Never). UFlowSettings exposes default policies as nullable const T* accessors. UFlowAsset holds a resolved policy instance with check() validation. 

Preload Helper (FFlowPreloadHelper / FFlowPreloadHelper_Standard)
New instanced-struct helper allocated per-node at InitializeInstance for any node (or addon) implementing IFlowPreloadableInterface. Tracks PendingPreloadCount (not a simple bool) so node + multiple addon participants are counted atomically before any PreloadContent call fires — prevents premature AllPreloadsComplete. Lifecycle hooks: OnNodeInitializeInstance, OnNodeActivate, OnNodeCleanup, OnNodeDeinitializeInstance, OnNodeExecuteInput. Safety flush on DeinitializeInstance regardless of flush timing policy. 

Context Pins 
Preloadable nodes gain Preload Content and Flush Content exec input pins and All Preloads Complete exec output pin automatically via GetContextInputs/GetContextOutputs. AllPreloadsComplete fires only when all participants (node + all preloadable addons) have reported completion.

AddOn Participation
TryInitializePreloadHelper allocates a helper when any addon implements IFlowPreloadableInterface, even if the node itself does not. TriggerPreload / TriggerFlush iterate preloadable addons alongside the node. Added UFlowNodeAddOn::NotifyPreloadComplete() (BlueprintCallable) — delegates to owning node, mirroring the Finish/TriggerOutput pattern.

Implementing Nodes
UFlowNode_PlayLevelSequence
— stores FStreamableHandle, binds CreateWeakLambda → NotifyPreloadComplete(), returns PreloadInProgress; FlushContent cancels the handle.
UFlowNode_ExecuteComponent
— delegates to component; PreloadInProgress from a component is guarded with ensureAlwaysMsgf (no callback path exists) and treated as Completed. UFlowNode_SubGraph
— synchronous CreateSubFlow, returns Completed.


MothDoctor:
Thank you for change and a few rounds of review.
Accepted :)

PR #362, opened 2026-04-03
Fix crash when OnMapOpened delegate fires after SLevelEditorFlow destruction

EvanC4:
The SLevelEditorFlow widget subscribes to FEditorDelegates::OnMapOpened using AddRaw, but never unsubscribes. When a map is loaded and the widget has been destroyed, the delegate fires on a stale pointer causing an access violation. Fix: Store the delegate handle and remove it in the destructor.


MothDoctor:
Oh... Thanks for patching this up. Fix integrated :)

PR #363, opened 2026-04-07
Fix FlowNode_LogicalOR: ExecutionLimit not working

Chen-Gary:
`FlowNode_LogicalOR` resets `ExecutionCount` to 0 every time when `Cleanup()`. Thus, `ExecutionCount` will never accumulate and never reach `ExecutionLimit`. Essentially, if we set `ExecutionLimit` to any value greater than 1, `FlowNode_LogicalOR` becomes an always-trigger node (i.e., can trigger output indefinitely).

To make `ExecutionLimit` work, I removed `Cleanup()` override in `FlowNode_LogicalOR`.

image

BTW, I didn't see many use cases of having `ExecutionLimit`. If users need a counter-like FLowNode, `UFlowNode_Counter` is a better choice.

### DIFF
--- a/Source/Flow/Private/Nodes/Route/FlowNode_LogicalOR.cpp
+++ b/Source/Flow/Private/Nodes/Route/FlowNode_LogicalOR.cpp
@@ -53,12 +53,19 @@ void UFlowNode_LogicalOR::ExecuteInput(const FName& PinName)
 	}
 }
 
-void UFlowNode_LogicalOR::Cleanup()
-{
-	ResetCounter();
-}
-
 void UFlowNode_LogicalOR::ResetCounter()
 {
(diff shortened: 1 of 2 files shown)


MothDoctor:
> If users need a counter-like FLowNode, UFlowNode_Counter is a better choice.
Conceptually, yes. Although the current implementation of the Counter node doesn't allow automatic disabling of the node once the counter is "completed". So it can be used many times.

IIRC `FlowNode_LogicalOR` is more often used in a narrative context ("player this action or another"), while the Counter node is more useful in repeatable gameplay actions (like what NPCs are doing when you run around them).

However, this implementation rose from my experience. It might be worth revisiting these nodes in the future. For now, thanks for the fix! :)

PR #365, opened 2026-04-11
CanSave check in FlowComponent

Bargestt:
Always wanted this check, not sure why it was never added to the plugin
It provides easy extensions if you don't want to clutter your save game
1) Option to implement save optimizations like Dirty flags
2) Transient graphs and components that are never saved. 
   - environment background behavior
   - logic/fact ensures
   - level initialization
   
And while transient graphs can be handled by using empty identity, it would be much better to intentionally set them up without looking like an error


MothDoctor:
Interesting, I never thought about using Flow graphs as some transient graphs. It's a simple change, so accepted :)

PR #366, opened 2026-04-11
GetLoadedRecord adjust interface

Bargestt:
In my opinion, this accessor will be much easier to extend, because when you need to override these functions you probably changed how data stored, but not at stage where you need rework whole subsystem

For example this was enough to create derived subsystem that stores data per SubLevel/WorldPartition Cell and applies Save/Load with streaming


MothDoctor:
That's indeed an elegant change. Accepted already :)

PR #367, opened 2026-04-11
Fix MatchAll with NotExact check

Bargestt:
FindComponents with filter MatchAll ignores bExactMatch and always checks Exact tag
I'm not sure how intended is this, or maybe it works as intended and I'm seeing things?
Anyways,if it is indeed a bug, here is the fix

Here is example of the problem I'm talking about

```
// This one matches all flow components on scene. Works as intended
Tags: Flow | Type: Any, Exact: false
  Cube_Flow_1                    Tags: Flow.Object.1
  Cube_Flow_12B                  Tags: Flow.Object.1, Flow.Object.2, Flow.Modifier.B
  Cube_Flow_1A                   Tags: Flow.Object.1, Flow.Modifier.A
  
// Exact match both tags. Works as intended
Tags: Flow.Object.1, Flow.Modifier.A | Type: All, Exact: true
   Cube_Flow_1A                   Tags: Flow.Object.1, Flow.Modifier.A

// Match Any. Works as intended
Tags: Flow.Object | Type: Any, Exact: false
  Cube_Flow_1                    Tags: Flow.Object.1
  Cube_Flow_12B                  Tags: Flow.Object.1, Flow.Object.2, Flow.Modifier.B
  Cube_Flow_1A                   Tags: Flow.Object.1, Flow.Modifier.A

// This one should have been identical to previous because we are matching using 1 tag
Match. Tags: Flow.Object Type: All Exact: false

```

Editor example. 
OnActorRegistered MatchAll should have triggered with selected cube, but it didn't
image

### DIFF
--- a/Source/Flow/Private/FlowSubsystem.cpp
+++ b/Source/Flow/Private/FlowSubsystem.cpp
@@ -708,11 +708,13 @@ void UFlowSubsystem::FindComponents(const FGameplayTagContainer& Tags, const EGa
 			TArray<TWeakObjectPtr<UFlowComponent>> ComponentsPerTag;
 			FindComponents(Tag, bExactMatch, ComponentsPerTag);
 			ComponentsWithAnyTag.Append(ComponentsPerTag);
+			break;
 		}
 
 		for (const TWeakObjectPtr<UFlowComponent>& Component : ComponentsWithAnyTag)
 		{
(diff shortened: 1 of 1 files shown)


MothDoctor:
It seems that's a bug, and it probably stayed hidden for a long time. Thank you!

PR #369, opened 2026-04-13
Actor details can show flow tags in selected actors

Bargestt:
Convenience access to flow component tags from selected actor details
Also makes possible to edit tags when multiple actors selected
Can be disabled from settings
image
image


Bargestt:
Found some issues with the widget.
It breaks interactive drag for sliders. Something with using External property in details
No idea how to fix it while keeping multiple values mode available in tag widget

Bargestt:
Finally found an overcomplicated way to avoid issue with broken sliders.

PR #376, opened 2026-04-27
Editor only pin text and tooltip

Bargestt:
Wrapped editor only properties in FFlowPin
Otherwise it generates absurd amount of duplicate entries in localization, I'm talking about thousands of entries in big projects

Shouldn't cause problems with migrating nodes if pins are created in Editor functions. Setting up pins in constructor is a little more finicky now though.


MothDoctor:
Oh, this nice catch and easy fix. I should integrate a long time ago, but that's finally done.
Thank you! :)

PR #378, opened 2026-06-02
Flow subgraph outputs and sequencer

LindyHopperGT:
- Flow Subgraph Output Data Pin support
- Sequencer fixes (particularly with regard to networking)
- FinishFlow changes (subgraphs not discarded until DeinitializeInstance, rather than OnCleanup, so they match the same cadence for deallocation)


MothDoctor:
Closing omni-PR as finishing integration.

PR #379, opened 2026-06-03
Subgraph deallocating timing change

LindyHopperGT:
Subgraph nodes were the only flow node that was permanently disposing of instance data on Cleanup as opposed to in DeinitializeInstance.  

Now subgraph nodes will cleanup their subgraph (including finishing it, etc.) but preserve the flow asset instance until the Subgraph flownode has its DeinitializeInstance called.


CyaDaPaKnat:
`virtual void FinishFlow(const EFlowFinishPolicy InFinishPolicy, const bool bRemoveInstance = true);`

This declaration was left in FlowAsset.h and it causes a compilation error on my end.

LindyHopperGT:
> `virtual void FinishFlow(const EFlowFinishPolicy InFinishPolicy, const bool bRemoveInstance = true);`
> 
> This declaration was left in FlowAsset.h and it causes a compilation error on my end.

this would be a typo.  

I split the omni PR into four focused PR's by hand without the ability to compile + test them locally (because our code relies on them, and I haven't made a silo flow for PR submission, which I really should).  

You can view the code in the omni change #378 to see what the code should be.  

If you could submit fixes to this PR, that would be very helpful.

CyaDaPaKnat:
> If you could submit fixes to this PR, that would be very helpful.

As it's my first time trying to contribute to an open-source project, I am not sure how to do that. From what I understood, I cannot directly push changes to the PR, did you maybe mean to suggest code changes ?

PR #380, opened 2026-06-03
DisplayName & category changes

LindyHopperGT:
For flow nodes that override GetNodeTitle/GetNodeDescription, only run those overrides for placed nodes in graphs. Preserve the default title and description for the archetype (e.g. in the node selection list)

Add flow node category  as BP native event for override in angel script


MothDoctor:
Change integrated with small tweak: `GetDisplayName()` renamed to `GetInstanceName()` for readability. And it's now virtual so now child classes can actually modify this :)

PR #381, opened 2026-06-03
Flow Sequencer changes

LindyHopperGT:
Adds UMKTFlowNodeAddOn_LevelSequenceActorBinding, a Flow node add-on that binds a named Level Sequence actor slot to an actor resolved from a blackboard key before playback. Supports multiplayer via replicated binding state on AFlowLevelSequenceActor. New IMKTPlayLevelSequenceAddOnInterface: pre-play hook (OnSequencePlayerCreated) for add-ons to apply binding overrides after player creation MKTFlowNode_PlayLevelSequenceAtStageMarker: accepts IMKTPlayLevelSequenceAddOnInterface children, calls the hook, exposes GetSequenceActor() AFlowLevelSequenceActor: AddBinding()/ClearAllBindings() with replicated BindingEntries so binding overrides propagate to clients


MothDoctor:
Nice! It's one of these multiplayer things I never time to handle in the open-source implementation :)

PR #382, opened 2026-06-03
subgraph data pin outputs

LindyHopperGT:
[Flow] Subgraph data pin outputs
Added subgraph data pin outputs support to Flow
Define the outputs UFlowAsset::OutputDataPinDeclarations The outputs are populated as input pins on Finish and SetOutputValue flownodes The outputs of the subgraph are populated as output pins for subgraph nodes The outputs for a subgraph execution are cached in the subgraph node.


MothDoctor:
Integrated to, likely, big joy among users :)

Submitted changes on my side
- Added clearing the `CachedOutputDataPinValues` while starting the Sub Graph so we don't use data from the previous asset instance on this node.
- Standard cosmetic pass. Added some comments inside of functions for readability of people try to understand lifetime of Flow by reading the code.

I flirted with making output data saveable. Already got a code, but I will push once I'd got some time to test it properly.

PR #385, opened 2026-06-19
Fixed CreateSubFlow returning nullptr if instance was preloaded

CyaDaPaKnat:
This fixes a problem where nullptr is returned from `CreateSubFlow` in `LoadSubFlow`.


MothDoctor:
Thanks for a fix :)

PR #387, opened 2026-07-11
Correctly supply data pins on packaged builds for subgraph node

dyanikoglu:
`TrySupplyDataPin` override in `UFlowNode_SubGraph` was inside `WITH_EDITOR` macro, which causes packaged builds to not properly work with subgraph flow asset params.


MothDoctor:
Oh, that lurked in shadows for some time now. Thanks for the fix!

PR #388, opened 2026-07-22
Fix: -Wunreachable-code-loop-increment errors on Clang

northstarswap:
Under Clang 20.1.8 on UE 5.8.0, `-Wunreachable-code-loop-increment` is set to `-Werror`. This causes loops that unconditionally break/return to fail to compile.

This PR reworks instances of these unconditional loops in `FlowSubsystem.cpp` and `FlowGraphEditor.cpp` so that the plugin compiles under Clang, as well as adds nullptr sanity checks where applicable.


MothDoctor:
Oh, that's interesting flag :)

I did modify your PR, but you did no mistake in your commit. I noticed shortcoming in the old code while reviewing your change. Code in the current shape satisfies Clang need for conditional exit from the loop. I tested that by compiling sample project with `-Compiler=Clang -UnreachableCodeWarningLevel=Error -NoLink`.

Thank you for the change!

PR #391, opened 2026-09-08
Added NotifyMatchType to OnNotifyFromActor

fede-Raider:
## Summary
Adds a new `NotifyMatchType` (EFlowTagMatchType) property to `UFlowNode_OnNotifyFromActor`, letting users choose how the NotifyTag should be matched.

## Motivation
Previously, `OnNotifyFromComponent` matched NotifyTags using `NotifyTags.HasTagExact(Tag)`, which only accepted exact tag matches. 
This made it impossible to set up a node listening for a parent tag (e.g. `Notify.Interaction`) that should trigger on any of its children (e.g. `Notify.Interaction.Pickup`, `Notify.Interaction.Use`).

## Changes
- Introduced `EFlowTagMatchType` type.
- Added `EFlowTagMatchType NotifyMatchType` property to `UFlowNode_OnNotifyFromActor`.
- Replaced `NotifyTags.HasTagExact(Tag)` with `FlowTypes::HasMatchingTag(Tag, NotifyTags, NotifyMatchType)`, mirroring the existing pattern already used for `IdentityMatchType`.

## Testing
- Verified exact-match behavior is unchanged when `NotifyMatchType` is set to `HasExact` (default value).
- Verified parent tag now correctly matches child tags when `NotifyMatchType` is set to `Has`.


Source/Flow/Private/Nodes/Actor/FlowNode_OnNotifyFromActor.cpp
```diff
@@ -33,27 +37,14 @@ void UFlowNode_OnNotifyFromActor::ForgetActor(TWeakObjectPtr<AActor> Actor, TWea
...
-		case EFlowTagContainerMatchType::HasAll:
-			IdentityMatches = Component->IdentityTags.HasAll(IdentityTags);
-			break;
-		case EFlowTagContainerMatchType::HasAllExact:
-			IdentityMatches = Component->IdentityTags.HasAllExact(IdentityTags);
-			break;
-	}
-
-	if (IdentityMatches && (!NotifyTags.IsValid() || NotifyTags.HasTagExact(Tag)))
+	const bool IdentityMatches = FlowTypes::HasMatchingTags(Component->IdentityTags, IdentityTags, IdentityMatchType);
```
fede-Raider, line 40: Use of already existing utility function

PR #392, opened 2026-09-10
Expose node input/output pins getters, and pin type name to blueprint to enable auto testing

omarchuk-gsc:
Hello, we have taken some early steps with automated testing, but we lack some functionality in the blueprints space. 
Changes are intended to be minimal and harmless.


MothDoctor:
Change integrated, thanks! :)

Issue #384, opened 2026-06-15
Incorrect World context during ForcePinActivation


CyaDaPaKnat:
I managed to reproduce the issue on the [FlowGame](https://github.com/MothCocoon/FlowGame) demo project.

In the editor, the steps to reproduce are :

1. Start PIE
2. Open the `q01_FirstActions` asset
3. Force the activation of the `Spawn` pin on the `Spawn by Actor Reference` node
4. Notice how no cube was spawned in the PIE level
5. Exit PIE
6. Notice how the cube was actually spawned in the editor level

PR #394, opened 2026-09-22
Added PIE world context guard during ForcePinActivation

CyaDaPaKnat:
We need to manually override the WorldContext which also sets the PlayInEditorID during a forced ExecuteInput call.

This ensures that soft pointers will resolve to the correct object or that actors will not be spawned into the Editor world during PIE.


CyaDaPaKnat:
This is tied to the [issue](https://github.com/MothCocoon/FlowGraph/issues/384) I previously opened.

